PR #892, opened 2014-07-21
Add: Documentation for local user inventory

nickanderson:
Parses /etc/shadow and /etc/passwd and report on discovered users and
attributes. Additional bundles available for augmenting Mission Portal
Inventory UI.

Ref: https://dev.cfengine.com/issues/6213


kacfengine:
@nickanderson: Can we close this, now that https://github.com/cfengine/masterfiles/pull/292 has been published separately?

PR #978, opened 2015-01-12
No class based decisions inside file control bodies.

rsevero:
Just making clear that class based decisions are not allowed inside file control bodies.

### DIFF
--- a/reference/components/file_control_promises.markdown
+++ b/reference/components/file_control_promises.markdown
@@ -24,6 +24,7 @@ tags: [body, bodies, components, common, namespace, promises, bundlesequence]
 This directive can be given multiple times within any file,
 outside of body and bundle definitions.
 
+Class based decisions are not allowed inside `file control` bodies.
 
 ### inputs
 


kacfengine:
I don't believe this is true. You should be able to use hard classes (like ``linux``) inside file control bodies. Are you saying that doesn't work?

rsevero:
On Wed, Jan 14, 2015 at 6:33 AM, Kristian Amlie
<notifications@github.com> wrote:
> I don't believe this is true. You should be able to use hard classes (like
> linux) inside file control bodies. Are you saying that doesn't work?

Humm... I tried soft classes and they don't work.

Never tried hard classes. Just tested. They actually work.

I will fix the message and resend, ok?


Rodrigo

kacfengine:
@rsevero: I'd still like to know whether it is soft classes in general or just because of ordering.

@pasinskim: If anything doesn't work here, it is surely because of the order that classes get set. In your experience with setting classes in the pre-evaluation stage, is any soft class expected to work in a file control body?

kacfengine:
@pasinskim: ping

kacfengine:
Clarified with https://github.com/cfengine/documentation/commit/7ce66210b83a71b491621dd0a3ad5e5337516755.

PR #979, opened 2015-01-12
Change: Fix api docs for host operations

nickanderson:
The host id includes SHA= in the front. API commands need to include that as
well or the queries will give incorrect results.

### DIFF
--- a/examples/enterprise-api-examples/browsing-host-information.markdown
+++ b/examples/enterprise-api-examples/browsing-host-information.markdown
@@ -112,8 +112,10 @@ split and indented for presentability).
 
 **Request**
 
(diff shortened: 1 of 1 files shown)


nickanderson:
@maciejmrowiec @estenberg 

This should be picked back to 3.6 as well.

estenberg:
Good catch. I think this changed from 3.5 to 3.6 unfortunately. Hopefully there aren't too many other places with this problem.

Thanks!

estenberg:
Cherry-picked to 3.6 with 35d664e2273b3e4f125be74b5356eaafdd5ac435

PR #993, opened 2015-02-05
Fix type 'atl_class' instead of 'alt_class'.

elentar:


elentar:
Just noticed a typo and figured I'd submit a correction.

kacfengine:
Thanks!

I merged manually to the master branch and 3.6. We don't build the 3.5 documentation anymore, so I left that one untouched.

PR #996, opened 2015-02-24
Add report_diffs limitations detailed description.

pasinskim:


pasinskim:
@nickanderson @estenberg please check if this is enough.

estenberg:
Looks good!

The 4500-line-limit between first and last change is very strange from the user's perspective. What can the user do about this? Ensure the file is changed in a "nice" way?
It would be much better to say e.g. that maximum file size is 2MB and cover all of this, but that's not the case.

Anyway, this is great progress, good enough for now. Merging and cherry-picking to 3.6.

estenberg:
Cherry-pick 3.6: ad987fa2db57dd3764f8ba9d5a39fbbeb8f9b81f

PR #998, opened 2015-03-04
Add: How to pass a datacontainer

nickanderson:
This is a frequently asked question.

This should also be picked into 3.6.x


kacfengine:
I learned something too! :-)

Cherry-picked to 3.6.

PR #1008, opened 2015-03-21
data_regextract: document new function

tzz:


tzz:
as requested in https://github.com/cfengine/core/pull/2134

PR #1009, opened 2015-03-27
Copied in Evolve Thinking's Mustache example, with permission.

neilhwatson:


kacfengine:
Merged with some minor modifications. Thanks Neil!

PR #1020, opened 2015-04-21
Redmine#2504: document variable string contexts

tzz:
As requested in https://github.com/cfengine/core/pull/2150


kacfengine:
Thank you!

PR #1021, opened 2015-04-24
Redmine#7111: @if macro

tzz:
Requested in https://github.com/cfengine/core/pull/2176


tzz:
I am anticipating that `@if` will be backported to 3.6.6 as we discussed in that PR.

PR #1022, opened 2015-04-24
reference/functions/file_hash.markdown: add docs

tzz:
Docs for the `file_hash()` function.  Require https://github.com/cfengine/core/pull/2189 for the example.  Should be backported to 3.6.


kacfengine:
Merged! Thanks for writing the docs, Ted. ``file_hash`` is not present in 3.6 though, as you indicated. At least I cannot find a reference to it in the code, so I changed the History section of the function reference to 3.7.0.

PR #1029, opened 2015-05-07
Change: Index on Benchmark table

nickanderson:
This is a better index per Maciej. Also updated the markdown to render using
the console brush for a (hopefully) nicer look.


nickanderson:
@estenberg @maciejmrowiec if the build renders OK at then please cherry pick this to 3.6 branch.

estenberg:
Looks good.

kacfengine:
Doc build is broken, error messages are always terrible, but I think it may be because of the masterfiles reorganization. It looks like it's a nontrivial task to fix, so I won't focus on it right now.

PR #1035, opened 2015-05-18
Document the hub's need for uname -n and maxconnection if it has many clients

ediosyncratic:


guide/installation-and-configuration/general-installation/installation-enterprise.markdown
```diff
@@ -94,10 +94,18 @@ The disk that serves PostgreSQL (/var/cfengine/state/pg) should be able to perfo
 
 **Note** Your storage IOPS specification may be given in 4KiB block size, in which case you would need to divide it by 4 to get the corresponding 16KiB *theoretical maximum*.
 
+### Open file descriptors
+
+The policy server should ideally be able to accept connections from all clients; i.e. to allow at least as many incoming connections as there are clients.
+The system limit for this is controlled by `ulimit -n`; so the parent process from which you bootstrap should, for a server with 5000 hosts, run `uname -n 5000` first.
+You should also add such a `uname -n` command to the script that implements `service cfengine3 start` (and `restart`).
+For very large numbers of clients, it may be advantageous to build a custom kernel to allow setting `uname -n` high enough.
```
estenberg: I would leave this out; as kernel builds nowdays have a high enough ulimit (>5000)?

guide/installation-and-configuration/general-installation/installation-enterprise.markdown
```diff
@@ -94,10 +94,18 @@ The disk that serves PostgreSQL (/var/cfengine/state/pg) should be able to perfo
 
 **Note** Your storage IOPS specification may be given in 4KiB block size, in which case you would need to divide it by 4 to get the corresponding 16KiB *theoretical maximum*.
 
+### Open file descriptors
+
+The policy server should ideally be able to accept connections from all clients; i.e. to allow at least as many incoming connections as there are clients.
+The system limit for this is controlled by `ulimit -n`; so the parent process from which you bootstrap should, for a server with 5000 hosts, run `uname -n 5000` first.
```
estenberg: ulimit -n 5000, not uname.

guide/installation-and-configuration/general-installation/installation-enterprise.markdown
```diff
@@ -94,10 +94,18 @@ The disk that serves PostgreSQL (/var/cfengine/state/pg) should be able to perfo
...
+You should also add such a `uname -n` command to the script that implements `service cfengine3 start` (and `restart`).
```
estenberg: Not uname.. Also, is it possible to add this to the CFE policy for large hubs (or is it tied to the shell or something else)? @nickanderson would know where it would make sense (probably somewhere in cfe_internal).

guide/installation-and-configuration/general-installation/installation-enterprise.markdown
ediosyncratic: For some values of "very large", however, that won't be enough.

guide/installation-and-configuration/general-installation/installation-enterprise.markdown
ediosyncratic: d'oh !  Thanks for spotting that.  Fixed in all three places.

ediosyncratic:
@estenberg said (on a comment that just went away because I pushed a fix to my dumb typo):
> [...] is it possible to add this to the CFE policy for large hubs (or is it tied to the shell or something else)? @nickanderson would know where it would make sense (probably somewhere in cfe_internal).

It needs to be run in an ancestor process of whatever starts cf-serverd and cf-hub (so running it in cf-execd would mostly suffice).  Anything you run from a child process only affects it and its descendants.  I don't know if policy has a way to ask us to do the relevant system call: I don't know what that syscall would be.

nickanderson:
The kernel default for maximum number of file descriptors (ulimit -n) was 1024/1024 (soft, hard), and has been raised to 1024/4096 in Linux 2.6.39. (http://serverfault.com/questions/356962/where-are-the-default-ulimit-values-set-linux-centos)

Should we suggest setting it in /etc/security/limits.conf?
```
root hard nproc <value>
root soft nproc <value>
```

nickanderson:
That would apply system wide to root, instead of something set specifically from cf-execd.

ediosyncratic:
I think documenting it as I have done makes more sense than setting the value globally.
Given that we do talk about networks with 5k clients, which is more than 4096, I think the comment about a custom kernel is apt.
Note that I took care to keep all the language "it'll work better this way" rather than "don't expect it to work without this"; we probably cope adequately without.

It might make sense for our service script to included a commented-out
```
# Uncomment this and set the limit to the number of clients in your network
# ulimit -n 1234
```
somewhere in the code that starts the service.

One thing I haven't documented is the collect_window setting, that I have in another pull request against masterfiles, albeit pending a bug-fix elsewhere.

guide/installation-and-configuration/general-installation/installation-enterprise.markdown
```diff
@@ -94,10 +94,18 @@ The disk that serves PostgreSQL (/var/cfengine/state/pg) should be able to perfo
 
 **Note** Your storage IOPS specification may be given in 4KiB block size, in which case you would need to divide it by 4 to get the corresponding 16KiB *theoretical maximum*.
 
+### Open file descriptors
+
+The policy server should ideally be able to accept connections from all clients; i.e. to allow at least as many incoming connections as there are clients.
```
ediosyncratic: Should I add a line after this saying, roughly, "This isn't essential: the system should work acceptably without it, but it may work significantly better with it." ?

estenberg: I don't think that's necessary. You already say "..should ideally..", which I think is sufficient.

estenberg:
Cherry-picked to 3.7 and 3.6 as well.

PR #1038, opened 2015-05-19
Redmine#7166: data_expand() documentation

tzz:
As requested in https://github.com/cfengine/core/pull/2207


kacfengine:
Thanks, Ted!

PR #1041, opened 2015-06-03
Fix broken instructions on how to set up git config.

ediosyncratic:
You need to run the git config commands as cfapache; and the git
config commands to set user.email and user.name should be passed the
--global flag.


kacfengine:
Cherry-picked to 3.7.

PR #1054, opened 2015-07-06
Redmine #7140: Add documentation for new package promise.

kacfengine:


kacfengine:
Merging immediately, in order to have it built and tested, but feel free to comment.

PR #1066, opened 2015-07-09
Change: Update Policy framework guide for 3.7

nickanderson:


nickanderson:
@tzz Do you want to add anything specific to def.json?

tzz:
Nice and thorough.

nickanderson:
OK then lets see how it renders :-P

PR #1078, opened 2015-07-14
Fix: packagesmatching() example

nickanderson:
Now using example from core repository instead of inline text.
Ref: https://dev.cfengine.com/issues/7380


nickanderson:
Needs to goto 3.6.x, 3.7.x.

Depends on https://github.com/cfengine/core/pull/2279

PR #1079, opened 2015-07-14
Fix: Documenation about constant variables

nickanderson:
Constant variables have not worked in quite some time. Implementation fix was
attempted multiple times without success. It was decided to fix the
documentation.

Ref: https://dev.cfengine.com/issues/1492

This should go into 3.6, and 3.7


nickanderson:
@estenberg @pasinskim  for your review

estenberg:
Thanks! Backported to 3.6 and 3.7.

PR #1104, opened 2015-07-23
Cherry-pick api documentation updates

maciejmrowiec:


maciejmrowiec:
Excluding 2 most recent commits with changes docs, other changes can go even to 3.6 docs as this is fixing some parts of api docs.

PR #1142, opened 2015-09-09
Redmine#7152: Allow whitespace in class expressions

tzz:
see https://github.com/cfengine/core/pull/2255

### DIFF
--- a/guide/language-concepts/classes.markdown
+++ b/guide/language-concepts/classes.markdown
@@ -102,12 +102,29 @@ The easiest way to limit the application of a promise to certain conditions is t
 
        Evening::
          "Good evening!";
+
+       "! any"::
+         "This report won't ever be seen.";
+
+       # whitespace allowed only in 3.8 and later
+       Friday . Evening::
+         "It's Friday evening, TGIF!";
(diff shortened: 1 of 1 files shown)


guide/language-concepts/classes.markdown
```diff
@@ -225,8 +240,10 @@ how to make complex decisions in CFEngine, with readable results. It is like
 defining aliases for class combinations. Such class 'aliases' may be specified 
 in any kind of bundle.
 
-Note that whitespace is *not* allowed between operators, so something like
-`a . b`, though perhaps more readable, will be rejected by the CFEngine parser.
+Note that whitespace is **only** allowed between operators when the
+expression is quoted as a string, so something like `a . b`, though
+perhaps more readable, will be rejected by the CFEngine parser. You
+need to use `"a . b"` instead.
```
kacfengine: Didn't we change this so that it *is* allowed?

Can also remove the "when quoted as strings" part above in that case.

tzz:
Thanks for checking twice, and sorry for slipping on this one.  It's OK now, I think.

PR #1143, opened 2015-09-12
Secure bootstrap guide improvements.

jimis:


jimis:
It was already showing up, but somebody had to browse into the "installation and configuration" page. With these changes it should show in the main "guide" page.

BTW the doc pipeline is red again...

jimis:
If after merging the state of this guide is satisfactory, we should cherry-pick it to 3.6, and I should update it for 3.7, it should be much simpler there.

PR #1166, opened 2015-09-21
Redmine#4309: document body inheritance via inherit_from

tzz:
see https://github.com/cfengine/core/pull/2320#issuecomment-141394427

### DIFF
--- a/guide/language-concepts/bodies.markdown
+++ b/guide/language-concepts/bodies.markdown
@@ -55,6 +55,97 @@ The attributes within the body are then type specific. Bodies of type `perms` co
 
 Such bodies can be reused in multiple promises. Like bundles, bodies can have parameters. The body `mog` also consists of the file permissions, file owner, and file group, but the values of those attributes are passed in as parameters.
 
+#### Body Inheritance
+
+CFEngine 3.8 introduced body inheritance via the `inherit_from`
+attribute. It's a parameterized single-inheritance system, so a body
+can only inherit from one other body, and it can apply parameters. The
+two bodies **must** have the same type.
+
+Let's see it with the `system` and `mog` bodies from earlier:
+
+```cf3
+    body perms system
+    {
+      mode => "644";
+      owners => { "root" };
+      groups => { "root" };
+    }
+
+    body perms system_inherited
+    {
+      inherit_from => mog("644", "root", "root");
+    }
+
+```
+
+The earlier `system` body and this `system_inherited` body have the
+same effect, eventually applying mode `644`, owner `root`, and group
+`root`. But they are created differently: the first by explicitly
+listing the parameters; the other by applying parameters to the
+`inherit_from` chain of inheritance.
+
+Which one is better? Usually, inheriting from a more generic
+specification is considered a better design pattern because it reduces
+horizontal complexity. But it's less explicit and some users and sites
+will prefer a more explicit listing of body attributes and their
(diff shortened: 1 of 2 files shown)


guide/language-concepts/bodies.markdown
```diff
@@ -55,6 +55,57 @@ The attributes within the body are then type specific. Bodies of type `perms` co
...
+    body perms system
+    {
+      mode => "644";
+      owners => { "root" };
+      groups => { "root" };
+    }
+
+    body perms system_inherited
+    {
+      inherit_from => mog("644", "root", "root");
```
nickanderson: What if it has inherit_from => mog("644", "root", "root"); and then mode => 600 specified within the system_inherited body?

tzz: Latest in the inheritance chain wins.  I didn't think of that in the example, though.  Do you think it's worth including it (keeping in mind it's a beginner document) or should I put it elsewhere?

nickanderson: It just seems like the kind of question someone would ask. I would add an example for it, just to show that you can inherit attributes that are set in other bodies, and that you can override them.

guide/language-concepts/bodies.markdown
```diff
@@ -55,6 +55,57 @@ The attributes within the body are then type specific. Bodies of type `perms` co
 
 Such bodies can be reused in multiple promises. Like bundles, bodies can have parameters. The body `mog` also consists of the file permissions, file owner, and file group, but the values of those attributes are passed in as parameters.
 
+#### Body Inheritance
+
+CFEngine 3.8 introduced body inheritance via the `inherit_from`
+attribute. It's a parameterized single-inheritance system, so a body can only inherit from one other body, and it can apply parameters.
```
nickanderson: You say single-inheritance, but that might confuse someone so that they think you cannot inherit from a body that inherits from another body.

```cf3
body perms mode_system
{
  mode => "644";
}
body perms system
{
  inherit_from => mode_system;
  owner => "root";
  group => "root"
}
body perms system_inherited
{
  inherit_from => system;
}
```

nickanderson:
There should be documentation for the body attributes. Unfortunately I don't see that we have any way of easily documenting common body attributes (only bodies that are common to all promise types). So each attribute body will need its own blurb. I think that blurb should link back to the section in the language concepts.

tzz:
OK, I elaborated in the language concepts and added a new section to the common promise attributes.  I think that's a good place for it since there's only two of these.

I also added docs about the universal `meta` body attribute since I don't see docs for https://github.com/cfengine/core/pull/2088 anywhere.

guide/language-concepts/bodies.markdown
```diff
@@ -55,6 +55,97 @@ The attributes within the body are then type specific. Bodies of type `perms` co
...
+    }
+
+    body perms system_twice
+    {
+      inherit_from => system_once("mark");
+      mode => "646";
+    }
+
+```
+
```
nickanderson: Great!

reference/promise-types.markdown
```diff
@@ -34,6 +34,82 @@ depends on the [bundle][bundles] type:
 See each promise type's reference documentation for detailed lists of available
 attributes.
 
+## Common Body Attributes
```
nickanderson: Thanks!

PR #1167, opened 2015-09-21
Add: note about connection caching

nickanderson:


guide/introduction/networking.markdown
```diff
@@ -194,6 +194,9 @@ authentication. Secrets should not be transferred through policy, encrypted or
 not. Policy files should be considered public, and any leakage should not
 reveal secret information.
 
+**Note:** Connections from the cf-agent are cached as described in the
+documnetation for body [`copy_from`][files#copy_from].
```
jimis: s/documnetation/documentation/

reference/promise-types/files.markdown
```diff
@@ -681,6 +681,12 @@ Diffs for binary files are not generated. Files are considered binary files if [
 
 **Type:** `body copy_from`
 
+The copy_from body specifies the details for making remote copies.
+
+**Note:** Connections from the cf-agent to cf-serverd do not persist beyond a
+single agent run. Connections may be re-initialized for each bundle that uses a
+remote connection.
+
```
jimis: Thanks Nick!

Personally I would try to mention it in a positive way, to present connection caching as a feature. "Connections from cf-agent to cf-serverd on the same host, get re-used for performance reasons. Implementation details currently restrict this connection caching only within the same bundle."

PR #1176, opened 2015-09-22
Add: Reference to common body attributes from bodies

nickanderson:


nickanderson:
@tzz a followup to your common body attributes.

PR #1183, opened 2015-09-29
get-package-data passes File as input, not Name

bahamat:
Update the doc to be consistent with actual cfengine behavior.


kacfengine:
Thanks! Easy to confuse the two.

kacfengine:
Cherry-picked to 3.7 as well.

PR #1189, opened 2015-10-19
Change: Relocate lanaguage concepts to reference

nickanderson:
- As discussed during the 2015 Q4 Community Advisory Board Meeting

- Ref: https://github.com/cfengine/cfengine-community-advisory-board/blob/master/minutes/2015-10-14.org#moving-language-concepts-guide-to-reference-section-of-documentation


nickanderson:
THIS IS AN EXPIRIMENT

I want to see if it's as easy as simply moving the content. If this fails badly then I will revert it.

PR #1201, opened 2015-11-09
typo

jimis:


jimis:
@nickanderson AFAIK there is no "logstring", maybe this is a historical remnant?

nickanderson:
Right, I see other examples showing log_string, and the reference shows log_string as well.
I think your right:

https://docs.cfengine.com/docs/3.7/reference-promise-types.html#log_string

PR #1224, opened 2015-11-11
Add: Historical default value for collect_window

nickanderson:
This default value was changed for performance reasons.


nickanderson:
@estenberg I added the prior default to the History section.

estenberg:
Thank you.

PR #1230, opened 2015-11-12
Fix: host API context query parameters

nickanderson:
They were inverted, examples showed the correct usage.


nickanderson:
This is already correct in 3.8 and master

PR #1234, opened 2015-11-18
Add docs about file_type option added in cfengine/core#2194

worr:
Pending merge of cfengine/core#2194, this change adds documentation for the ability to create fifos through policy.


kacfengine:
Thanks!

PR #1235, opened 2015-11-20
Redmine #7700 Fix upgrade procedure for 3.8

pasinskim:
Changelog: Updated update procedure for CFEngine 3.8.


pasinskim:
@nickanderson @kacfengine now the order should be as expected.

PR #1236, opened 2015-11-20
Change: Improve reports type promise docs

nickanderson:
- Include more complete example from core

Depends on example in https://github.com/cfengine/core/pull/2411

### DIFF
--- a/reference/promise-types/reports.markdown
+++ b/reference/promise-types/reports.markdown
@@ -6,19 +6,18 @@ tags: [reference, bundle common, reports, promises]
 ---
 
 Reports promises simply print messages. Outputting a message without
-qualification can be a dangerous operation. In a large installation it could
-unleash an avalanche of messaging.
+qualification can be a dangerous operation. In a large installation it
+could unleash an avalanche of messaging, so it is reccomended that reports are
+appropriately.
 
-```cf3
-    reports:
+[%CFEngine_include_example(reports.cf)%]
 
-    "literal string or file refererence",
-       printfile = printfile_body,
-       ...;
-```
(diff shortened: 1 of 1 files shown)


reference/promise-types/reports.markdown
```diff
@@ -6,19 +6,18 @@ tags: [reference, bundle common, reports, promises]
 ---
 
 Reports promises simply print messages. Outputting a message without
-qualification can be a dangerous operation. In a large installation it could
-unleash an avalanche of messaging.
+qualification can be a dangerous operation. In a large installation it
+could unleash an avalanche of messaging, so it is reccomended that reports are
+appropriately.
```
kacfengine: `s/reccomended/recommended/`. Also the last sentence ends abruptly, and doesn't make sense.

reference/promise-types/reports.markdown
```diff
@@ -6,19 +6,18 @@ tags: [reference, bundle common, reports, promises]
...
-    reports:
+[%CFEngine_include_example(reports.cf)%]
 
-    "literal string or file refererence",
-       printfile = printfile_body,
-       ...;
-```
+{% comment %} TODO: Should link to a page that describes all cfengine output
+here where we explain that output from reports type promises are
+prefixed with the letter R. {% endcomment %}
```
kacfengine: Didn't know about this notation. This will cause it not to appear in the documentation rendering?

reference/promise-types/reports.markdown
nickanderson: Thank you

reference/promise-types/reports.markdown
nickanderson: Yep, I keep finding little gems

PR #1254, opened 2015-12-07
Redmine#7633: processexists() and findprocesses()

tzz:
@pasinskim as requested in https://github.com/cfengine/core/pull/2349

### DIFF
--- a/reference/functions/findprocesses.markdown
+++ b/reference/functions/findprocesses.markdown
@@ -1,0 +1,42 @@
+---
+layout: default
+title: findprocesses
+published: true
+tags: [reference, process functions, functions, findprocesses, process, processes, ps]
+---
+
+[%CFEngine_function_prototype(regex)%]
+
+**Description:** Return the list of processes that match the given regular expression `regex`.
+
+This function searches for the given regular expression in the process
+table. Use `.*sherlock.*` to find all the processes that match
+`sherlock`. Use `.*\bsherlock\b.*` to exclude partial matches like
+`sherlock123` (`\b` matches a word boundary).
+
+[%CFEngine_function_attributes(regex)%]
+
+The returned data container is a list of key-value maps. Each one is
+guaranteed to have the key `pid` with the process ID. The key `line`
+will also be available with the raw process table contents.
+
+The process table is usually obtained with something like
+`ps -eo user,pid,ppid,pgid,%cpu,%mem,vsize,ni,rss,stat,nlwp,stime,time,args`
+but the exact data under the `line` key may change per platform and per CFEngine release.
+
+**Example:**
+
+```cf3
+    vars:
+      "holmes" data => findprocesses(".*sherlock.*");
+```
+
+Output:
+
+```
+    [ { "pid": "2378", "line": "...the ps output here" }, ... ]
+```
+
+**History:** Introduced in CFEngine 3.9
+
+**See also:** [`processes`][processes] [`processexists()`][processexists].
--- a/reference/functions/processexists.markdown
+++ b/reference/functions/processexists.markdown
@@ -1,0 +1,33 @@
+---
+layout: default
+title: processexists
+published: true
+tags: [reference, process functions, functions, processexists, process, processes, ps]
+---
+
+[%CFEngine_function_prototype(regex)%]
+
+ **Description:** Return whether a process matches the given regular expression `regex`.
+
+This function searches for the given regular expression in the process
+table. Use `.*sherlock.*` to find all the processes that match
+`sherlock`. Use `.*\bsherlock\b.*` to exclude partial matches like
+`sherlock123` (`\b` matches a word boundary).
+
+[%CFEngine_function_attributes(regex)%]
+
+The process table is usually obtained with something like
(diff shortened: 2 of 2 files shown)


tzz:
ping @pasinskim @kacf

reference/functions/findprocesses.markdown
```diff
@@ -0,0 +1,38 @@
...
+will also be available with the raw process table contents.
+
+The process table is usually obtained with something like
+`ps -eo user,pid,ppid,pgid,%cpu,%mem,vsize,ni,rss,stat,nlwp,stime,time,args`
+but the exact data under the `line` key may change per platform and per CFEngine release.
+
+**Example:**
+
+    vars:
+      "holmes" data => findprocesses(".*sherlock.*")
```
kacf: This should have a

```
```cf3
```.
```

around it, or? (pretend the last '.' isn't there, I had to avoid Github formatting)

kacf: Including the example:

```
```cf3
    vars:
      "holmes" data => findprocesses(".*sherlock.*")
```.
```

Again, ignore the last dot.

reference/functions/findprocesses.markdown
```diff
@@ -0,0 +1,38 @@
...
+
+Output:
+
+    [ { "pid": "2378", "line": "...the ps output here" }, ... ]
```
kacf: And this one should have the same, but without the "cf3" (I challenge you to render this stuff in Github!).

reference/functions/processexists.markdown
```diff
@@ -0,0 +1,38 @@
...
+but the exact data under the `line` key may change per platform and per CFEngine release.
+
+**Example:**
+
+    classes:
+      "holmes" expression => processexists(".*sherlock.*")
+
+Output:
+
+    [ { "pid": "2378", "line": "...the ps output here" }, ... ]
```
kacf: Same comments here regarding quoting the example and output.

reference/functions/findprocesses.markdown
```diff
@@ -0,0 +1,38 @@
+---
+layout: default
+title: findprocesses
+published: true
+tags: [reference, process functions, functions, findprocesses, process, processes, ps]
+---
+
+[%CFEngine_function_prototype(regex)%]
+
+**Description:** Return the list of files that match the given regular expression `regex`.
```
kacf: `s/list of files/list of processes/`

reference/functions/findprocesses.markdown
```diff
@@ -0,0 +1,38 @@
...
+
+[%CFEngine_function_attributes(regex)%]
+
+The returned data container is a list of key-value maps. Each one is
+guaranteed to have the key `pid` with the process ID. The key `line`
+will also be available with the raw process table contents.
+
+The process table is usually obtained with something like
+`ps -eo user,pid,ppid,pgid,%cpu,%mem,vsize,ni,rss,stat,nlwp,stime,time,args`
+but the exact data under the `line` key may change per platform and per CFEngine release.
```
kacf: It's not really clear from this text what will be available besides `pid` and `line`. Does it depend on the `ps` command line used?

reference/functions/processexists.markdown
kacf: Also, this example is wrong if I understand the function correctly. There is no returned data, just true/false.

reference/functions/processexists.markdown
```diff
@@ -0,0 +1,38 @@
...
+
+[%CFEngine_function_attributes(regex)%]
+
+The returned data container is a list of key-value maps. Each one is
+guaranteed to have the key `pid` with the process ID. The key `line`
+will also be available with the raw process table contents.
+
+The process table is usually obtained with something like
+`ps -eo user,pid,ppid,pgid,%cpu,%mem,vsize,ni,rss,stat,nlwp,stime,time,args`
+but the exact data under the `line` key may change per platform and per CFEngine release.
```
kacf: No returned data for this function, right? Just returns true or false.

reference/functions/findprocesses.markdown
tzz: It's an internal limitation: the code just returns an `ItemList` which is a key-value container. There is no more detail available. We could potentially parse `/proc` and get much more detail, but that's not possible currently.

@estenberg this detail would improve Enterprise reporting as well. With https://github.com/cfengine/core/pull/2245 we have a structured way of parsing data out of `/proc` that should make the task easy. But I don't know if it's possible to transmit data container JSON data to the Enterprise server, so I'm just mentioning it here FYI.

reference/functions/findprocesses.markdown
tzz: You have to use the "4 leading spaces make a code block" rule :)

    ```
    my output
    ```

reference/functions/processexists.markdown
tzz: Yup, fixed.

tzz:
All comments addressed. Sorry for the sloppy docs.

kacf:
Thanks for fixing!

PR #1264, opened 2015-12-18
fix a typo in sort() attributes

teroka:
Trying to use regex ends up badly


kacf:
Indeed, good catch, thanks!

kacf:
Cherry-picked to 3.6, 3.7 and 3.8 as well.

PR #1268, opened 2015-12-22
Make it clear that Python is required for new package modules.

kacf:
Redmine #7820.

### DIFF
--- a/reference/promise-types/packages.markdown
+++ b/reference/promise-types/packages.markdown
@@ -24,6 +24,9 @@ following package managers:
 * ``yum``: YUM package manager and accompanying rpm package manager.
 * ``apt_get``: Apt package manager and accompanying dpkg package manager.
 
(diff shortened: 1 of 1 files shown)


reference/promise-types/packages.markdown
```diff
@@ -24,6 +24,9 @@ following package managers:
 * ``yum``: YUM package manager and accompanying rpm package manager.
 * ``apt_get``: Apt package manager and accompanying dpkg package manager.
 
+Both ``yum`` and ``apt_get`` package managers require Python to be installed on
```
jimis: python 2 or 3? Which version and up, 2.7 or maybe 3.3?

kacf: I admit that I haven't tested with both, but I did use the recommended syntax which is supposed to work on both. Until we know of any problems I'll leave it unspecified, and if we find any, I think we should fix the code instead.

PR #1306, opened 2016-01-15
Allow some functions to take inline JSON: mapdata, mergedata, and others

tzz:
As requested in https://github.com/cfengine/core/pull/2454

This is a snapshot of the current state, putting all the docs in `mergedata()`. But when https://dev.cfengine.com/issues/7871 is done, we'll add a new `Inline JSON` section to the vars documentation and move all this information there. We can't do it yet because not all functions have been converted.


kacf:
I like "This is standard across many CFEngine functions", sounds really nice! :-)

PR #1308, opened 2016-01-22
Document the host properties used by select_class

skreuzer:
Instead of saying 'host identity' explicitly list out the 3 properties that are use to create the hash select_class will use to pick a class.


tzz:
That's good stuff, it should be backported all the way back to 3.6 docs IMO.

atsaloli:
Very nice!  I've always wondered how that works.  Thanks, skreuzer.

kacf:
Indeed, much clearer! Why cloud important information? :-)

kacf:
Cherry-picked to 3.6, 3.7 and 3.8 as well. Thanks!

PR #1310, opened 2016-01-25
Fixed typo in example cf file

dillongilmore:
I'm unsure if this PR is needed as I've never heard to CGEngine, but I'm also new to cfengine. I'm assuming this is  a typo as G is next to F and could be an easy mistake to make.


kacf:
It's a super secret doomsday device. Now that you've seen it, we may have to kill you!

... Or not. Thanks for spotting this! :-)

kacf:
Cherry-picked to 3.7 and 3.8 as well (3.6 doesn't have this example).

PR #1319, opened 2016-01-28
Update access_rules doc.

jimis:


jimis:
@nickanderson please review

PR #1320, opened 2016-02-12
Document function history

skreuzer:
Add a note on when the expandrange() function was added


kacf:
Thanks, Steven! Cherry-picked to 3.7 and 3.8 as well.

PR #1438, opened 2016-04-01
Document that files in suspiciousnames are skipped

amousset:
Linked to https://github.com/cfengine/core/pull/2548


jimis:
Thanks!

PR #1448, opened 2016-04-18
Add package_inventory note to packagesmatching and packageupdatesmatc…

pasinskim:
…hing.

Add description about package_inventory body common control attribute to
packagesmatching() and packageupdatesmatching() documentation.

### DIFF
--- a/reference/functions/packagesmatching.markdown
+++ b/reference/functions/packagesmatching.markdown
@@ -39,12 +39,24 @@ The return is a data container with a list of package descriptions, looking like
 The following code extracts just the package names, then looks for
 some desired packages, and finally reports if they are installed.
 
+**IMPORTANT:** Please note that you need to provide `package_inventory` attribute in `body common control` in order to be able to use this function. Also depending on the value(s) of `package_inventory` only packages from selected package modules will be returned. For more information about `package_inventory` please read [`package_inventory`][Components and Common Control#package_inventory] section.
+
 [%CFEngine_include_example(packagesmatching.cf)%]
 
 **Example:**
 
 ```cf3
       "all_packages" data => packagesmatching(".*", ".*", ".*", ".*");
+```
+
+**Refresh rules:** 
+* inastalled packages cache used by packagesmatching() is refreshed at the end of each agent run in accordance with constraints defined in the relevant package module body.
+* installed packages cache is refreshed after installing or removing a package.
+* installed packages cache is refreshed if no local cache exists.
+        This means a reliable way to force a refresh of CFEngine's internal package cache is to simply delete the local cache:
+
+```cf3
+            $(sys.statedir)/packages_installed_<package_module>.lmdb*
 ```
 
 **History:** Introduced in CFEngine 3.6
--- a/reference/functions/packageupdatesmatching.markdown
+++ b/reference/functions/packageupdatesmatching.markdown
@@ -34,11 +34,25 @@ The return is a data container with a list of package descriptions, looking like
 * `arch_regex` - Regular expression matching package architecutre
 * `method_regex` - Regular expression matching package method (apt-get, rpm, etc ...)
 
+**IMPORTANT:** Please note that you need to provide `package_inventory` attribute in `body common control` in order to be able to use this function. Also depending on the value(s) of `package_inventory` only packages from selected package modules will be returned. For more information about `package_inventory` please read [`package_inventory`][Components and Common Control#package_inventory] section.
+
+
 **Example:**
 
 ```cf3
       "all_package_updates" data => packageupdatesmatching(".*", ".*", ".*", ".*");
 ```
+
+**Refresh rules:**
+* updates cache used by packageupdatesmatching() is refreshed at the end of each agent run in accordance with constraints defined in the relevant package module body.
+* updates cache is refreshed every time `repo` type package is installed or removed
+* updates cache is refreshed if no local cache exists.
+        This means a reliable way to force a refresh of CFEngine's internal package cache is to simply delete the local cache:
+
+```cf3
+            $(sys.statedir)/packages_updates_<package_module>.lmdb*
+```
+
 
 **History:** Introduced in CFEngine 3.6
 


pasinskim:
@nickanderson can you check if this description makes sense?

nickanderson:
This indicates that you can only report for a single package manager. For example I would have to choose between pip installed packages vs RPM installed packages. Is that the case or is this just for default package inventory?

kacf:
`package_inventory` is a list, so you can query for whatever package managers are in the list.

pasinskim:
Right. The thing is that you need to fill the ```package_inventory``` list with the package modules you want to be reported. If the list is empty then ```packagesmatching()``` won't work.

kacf:
I suppose technically speaking it will work if you have package promises that use the same package module you're searching for, but this is not recommended, and I don't think we should even mention it. The official way is to use the `package_inventory`.

pasinskim:
@kacf actually it won't. We are querying database to get ONLY packages from modules matching entries in ```package_inventory```.

kacf:
Ok, I see. I thought these databases were shared.

kacf:
The notes seem good to me, btw.

reference/functions/packagesmatching.markdown
```diff
@@ -39,6 +39,8 @@ The return is a data container with a list of package descriptions, looking like
 The following code extracts just the package names, then looks for
 some desired packages, and finally reports if they are installed.
 
+*IMPORTANT:* Please note that you need to provide ```package_inventory``` attribute in ```body common control``` in order to be able to use this function. Also depending on the value(s) of ```package_inventory``` only packages from selected package modules will be returned. For more information about ```package_inventory``` please read [`package_inventory`][Components and Common Control#package_inventory] section.
```
nickanderson: Double asterisks for bolding in markdown.
**IMPORTANT:**

Single backticks for verbatium or automatic linking to cfengine attributes:

`package_inventory` `body common control`

pasinskim:
@nickanderson thanks for review. I've added ```refresh rules``` section describing when cache is updated for both functions.

reference/functions/packageupdatesmatching.markdown
```diff
@@ -34,12 +34,23 @@ The return is a data container with a list of package descriptions, looking like
 * `arch_regex` - Regular expression matching package architecutre
 * `method_regex` - Regular expression matching package method (apt-get, rpm, etc ...)
 
+**IMPORTANT:** Please note that you need to provide ```package_inventory``` attribute in ```body common control``` in order to be able to use this function. Also depending on the value(s) of ```package_inventory``` only packages from selected package modules will be returned. For more information about ```package_inventory``` please read [`package_inventory`][Components and Common Control#package_inventory] section.
```
nickanderson: single *`* througout

nickanderson:
looks pretty good, just replace the ```word``` with `word` except in the case of ```cf3 :)

reference/functions/packageupdatesmatching.markdown
```diff
@@ -34,12 +34,23 @@ The return is a data container with a list of package descriptions, looking like
...
 ```cf3
       "all_package_updates" data => packageupdatesmatching(".*", ".*", ".*", ".*");
 ```
 
+**Refresh rules:**
+* updates cache used by packageupdatesmatching() is refreshed at the end of each agent run in accordance with constraints defined in the relevant package module body.
+* updates cache is refreshed every time ```repo``` type package is installed or removed
+* updates cache is refreshed if no local cache exists.
+        This means a reliable way to force a refresh of CFEngine's internal package cache is to simply delete the local cache:
+            $(sys.statedir)/packages_updates_<package_module>.lmdb*
```
kacf: Quote file name to give it monospace font. Use triple backtick (`) notation without the cf3.

reference/functions/packageupdatesmatching.markdown
```diff
@@ -34,12 +34,23 @@ The return is a data container with a list of package descriptions, looking like
...
+**IMPORTANT:** Please note that you need to provide ```package_inventory``` attribute in ```body common control``` in order to be able to use this function. Also depending on the value(s) of ```package_inventory``` only packages from selected package modules will be returned. For more information about ```package_inventory``` please read [`package_inventory`][Components and Common Control#package_inventory] section.
+
+
 **Example:**
 
 ```cf3
       "all_package_updates" data => packageupdatesmatching(".*", ".*", ".*", ".*");
 ```
 
+**Refresh rules:**
```
kacf: Maybe also comment that network is only hit if the ifelapsed has expired, regardless of the rules below.

reference/functions/packagesmatching.markdown
```diff
@@ -47,6 +49,13 @@ some desired packages, and finally reports if they are installed.
       "all_packages" data => packagesmatching(".*", ".*", ".*", ".*");
 ```
 
+**Refresh rules:** 
+* inastalled packages cache used by packagesmatching() is refreshed at the end of each agent run in accordance with constraints defined in the relevant package module body.
+* installed packages cache is refreshed after installing or removing a package.
+* installed packages cache is refreshed if no local cache exists.
+        This means a reliable way to force a refresh of CFEngine's internal package cache is to simply delete the local cache:
+            $(sys.statedir)/packages_installed_<package_module>.lmdb*
```
kacf: Quote here as well, like the other one.

reference/functions/packageupdatesmatching.markdown
nickanderson: Why triple backtick? I use single backticks usually.

nickanderson: Oh you mean

backtickbacktickbacktickcf3
vs 
backtickbacktickbacktick

for verbatim blocks. yes. tick tick tick cf3 is for cfengine code. verbatim single words can be single backtick though.

kacf: Yes, I meant as a block, not as a word.

PR #1450, opened 2016-04-19
Redmine#7435: add docs for json_pipe integration

tzz:
As requested in https://github.com/cfengine/core/pull/2572 and https://github.com/cfengine/core/pull/2319

### DIFF
--- a/reference/functions/mapdata.markdown
+++ b/reference/functions/mapdata.markdown
@@ -11,7 +11,8 @@ tags: [reference, data functions, functions, mapdata, inline_json]
 array is a map across each element of `array_or_container`, modified by
 a `pattern`. The map is either collected literally when `interpretation`
 is `none`, or canonified when `interpretation` is `canonify`,
-or parsed as JSON when `interpretation` is `json`.
+or parsed as JSON when `interpretation` is `json`, or collected from `pattern`,
+invoked as a program, when `interpretation` is `json_pipe`.
 
 This is a [collecting function][Functions#collecting functions] so it can accept many types of data parameters.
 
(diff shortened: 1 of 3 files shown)


reference/functions/mapdata.markdown
```diff
@@ -36,6 +37,28 @@ Output:
...
+
+The `json_pipe` interpretation is intended to work with programs that take JSON
+as input and produce JSON as output. This is a standard tool convention in the
+Unix world. See the example below for the typical usage.
+
+[jq][https://stedolan.github.io/jq/] has a powerful programming language that
+fits the `json_pipe` interpretation well. It will take JSON input and product
+JSON output. Please read the [jq][https://stedolan.github.io/jq/] manual and
+cookbook to get a feel for the power of this tool. When available,
+[jq][https://stedolan.github.io/jq/] will offer tremendous data manipulation
```
nickanderson: **tremendous**

reference/special-variables.markdown
```diff
@@ -33,6 +33,8 @@ Variables defined in a monitoring context.
 * [sys][sys]
 Variables defined in order to automate discovery of system values.
 
+* [def][def]
+Variables with some default value that can be overridden by [augments file][Augments].
```
nickanderson: s/overridden/defined/

If you define something in the augments file, and you do not have special treatment in `bundle common def` it will be defined. But if you define a variable in `bundle common def` without a guard like `isvaribale()` then the variable as defined in policy wins.

reference/special-variables/def.markdown
```diff
@@ -31,3 +31,14 @@ bundle common def
   # ...
 }
 ```
+
+### def.jq
+
+This variable gives a convenient way to invoke
+[jq][https://stedolan.github.io/jq/] for the `mapdata()` function in `json_pipe`
+mode and elsewhere. Note the below is the **default** value that you can
```
nickanderson: Is this default in C or in `bundle common def` of the Masterfiles Policy Framework we should probably specify.

tzz:
@nickanderson I fixed your comments, thanks!

nickanderson:
looks good to me

kacf:
Nice!

kacf:
Merged manually.

PR #1452, opened 2016-04-20
Document callstack_callers() and callstack_promisers()

tzz:
as requested in https://github.com/cfengine/core/pull/2331#issuecomment-212341037 and https://github.com/cfengine/core/pull/2585 by @kacf


kacf:
Great, thanks!

PR #1453, opened 2016-04-20
Clarify JSON boolean handling and improve expression docs

tzz:
as requested by @jimis in https://github.com/cfengine/core/pull/2579

### DIFF
--- a/reference/language-concepts/classes.markdown
+++ b/reference/language-concepts/classes.markdown
@@ -304,6 +304,48 @@ unpredictable results are avoided. i.e negative knowledge does not necessarily
 imply that something is not the case, it could simply be unknown. Checking if
 each file exists before calling `isnewerthan` would avoid this problem.
 
+### Operands that are JSON booleans
+
+If an operand is `true` it will succeed, **even though there doesn't have to be
+a class named** `true`. If an operand is `false` it will fail, **even though
+there may be a class named** `false`. This allows JSON booleans from data
+containers to be used in context expressions:
+
+```cf3
+bundle agent main
+{
+    vars:
+      "checks" data => '[true, false]';
+      # find all classes named
+      "classes_named_true" slist => classesmatching('true');
+
+  classes:
+      # always defined
+      "first_check" expression => "$(checks[0])";
+      # never defined
+      "second_check" expression => "$(checks[1])";
+
+  reports:
+      # prints nothing, there are no classes named 'true'
+      "Classes named 'true': $(classes_named_true)";
+
+    first_check::
+      "The class was defined from '$(checks[0])'";
+    !first_check::
+      "The class was NOT defined from '$(checks[0])'";
+    second_check::
+      "The class was defined from '$(checks[1])'";
+    !second_check::
+      "The class was NOT defined from '$(checks[1])'";
+}
+```
+
+Output:
+
+```
+R: The class was defined from 'true'
+R: The class was NOT defined from 'false'
+```
 
 ## Global and Local classes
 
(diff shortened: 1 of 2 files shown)


jimis:
Looks good. Maybe also add a simple expression (i.e. no JSON) example? Maybe add it [here](https://docs.cfengine.com/docs/3.7/reference-promise-types-classes.html#expression) because I think that part is really empty, and the classes promise doc was the first one that came to my mind to look at. Or add a link toward the part you edited?

tzz:
@nickanderson @jimis @kacf that reminds me, what do you think about `classes: "myclass" ==> "any"`?

In other words, can we add a syntactic shortcut for the equivalent but much more verbose `classes: "myclass" expression => "any"`? The exact syntax is not so important, as long as it's very short and doesn't require parser changes.

nickanderson:
I'm not against it. But lets send this out to the community lists to see what others have to say.

nickanderson:
After that perhaps bring it up in the #CAB

tzz:
@nickanderson thank you for volunteering ;)

tzz:
@jimis I added docs as you requested. Please feel free to fix them or amend them after merge.

jimis:
> classes: "myclass" ==> "any"
I'm definitely **against**. No extra symbols, and this one specifically is rather confusing.

jimis:
Thanks, merging.

kacf:
>> classes: "myclass" ==> "any"
> 
> I'm definitely **against**. No extra symbols, and this one specifically is rather confusing.

I think I'm siding with Jimis on this one, it breaks with the whole language, and doesn't have a single keyword in it for recognition. Typing less is not always a good thing, as Perl has repeatedly shown (who thought `$|` or `$/` were good names?).

tzz:
Hmm, I think shell and awk syntax was the model for Perl, you have things like `$@` etc.

I think at least the common `classes: "x" expression => "any"` can be shortened to `classes: "x"`.

nickanderson:
I can get on board with classes: "any";

There is some precedence with methods: "method";

I also think that classes: "x" => "my.expression" is a decent candidate.

jimis:
> I also think that classes: `"x" => "my.expression"` is a decent candidate.

It /looks/ fine, but I will side with @zzamboni's opinion on the mailing list. I vote against because it will complicate the only clear thing in our syntax. :-)

zzamboni:
And just for reference, here is my comment from the mailing list:

> I’m not sure about this, and I think the reason is that I agree with Aleksey - one of the nice things of the CFEngine syntax is its regularity:
> 
> promise_type:
>    promiser
>       [ attribute => value … ]
> 
> The `attribute => value` pairs are optional - if we start making the `attribute` part optional, the syntax will get confusing (not to mention making the parsing more complicated). This is, if “expression” is optional for classes, what’s the default attribute for `vars`, for `files`, or for other promise types? While I can see the appeal in terms of shorter code, I fear it’s a slippery slope that could easily lead to confusing code:
> 
> files:
>    “/foo/bar” => og(“root”, “wheel”);   # “mode" by default?
> commands:
>   “/bin/foo” => “/etc”;      # “chdir” by default?
> etc.
> 
> I do like the part about `class: “foo”` being equivalent to `class: “foo” expression => “any”`, because it doesn’t break the syntax model.

nickanderson:
Right, count me in that boat.
I don't personally have a burning desire to shorten "foo" => "expression". 

The "class"; shortening has precedence with reports type promises as well as the methods short-cut as of 3.7.

PR #1457, opened 2016-04-22
Update classes.markdown

LINDEN-IT-NET:
Added 2x "," between "class" and scope =>.. because otherwise it doesn't work ;-)

### DIFF
--- a/reference/promise-types/classes.markdown
+++ b/reference/promise-types/classes.markdown
@@ -300,12 +300,12 @@ on [Negative Knowledge][classes and decisions].
 
 ```cf3
     classes:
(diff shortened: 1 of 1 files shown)


nickanderson:
Commas after the promiser (the quoted string at the beginning of a promise) are not valid syntax. 

But thank you for your willingness to contribute to the docs!

nickanderson:
That changed in 3.5 BTW

LINDEN-IT-NET:
But I had to put commas in my promise to get it working on 3.7.2.

Sure, I owe you some contribution ;-)

PR #1469, opened 2016-04-28
Change: Re-root edit_line promises under files promises

nickanderson:


nickanderson:
Well see what this looks like and revert it if necessary

nickanderson:
yeah it messed up some of the rendering for the left hand side navigation.

PR #1471, opened 2016-04-29
New feature: Classes promise: allow classes without an expression to default to defined.

tzz:
As requested in https://github.com/cfengine/core/pull/2596#issuecomment-215703224


jimis:
Thanks!

PR #1522, opened 2016-06-02
Fix incorrect syntax documentation for packages -> options.

kacf:
It is an slist, not a string.


kacf:
@pasinskim: Brought up by a community user.

kacf:
Cherry-picked to 3.7, 3.8 and 3.9 as well.

PR #1537, opened 2016-06-14
ENT-2714: Fix libltdl dependency.

jimis:


nickanderson:
beat me to it, thanks

PR #1539, opened 2016-06-16
Add: Note special behaviour change for ifelse()

nickanderson:
This needs picked to 3.7.x and 3.9.x

### DIFF
--- a/reference/functions/ifelse.markdown
+++ b/reference/functions/ifelse.markdown
@@ -92,8 +92,7 @@ bundle agent example
 
 **Note:** As a general rule function evaluation is skipped when undefined
 variables are used. However this function has special behavior when **exactly
-three** arguments are used allowing it to be actuated in order to enable the
-following use case.
+three** arguments are used, allowing it to be evaluated even if it contains undefined variables. For example:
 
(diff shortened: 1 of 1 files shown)


reference/functions/ifelse.markdown
```diff
@@ -89,3 +89,28 @@ bundle agent example
       "ifelse result list: $(mylist)";
 }
 ```
+
+**Note:** As a general rule function evaluation is skipped when undefined
+variables are used. However this function has special behavior when **exactly
+three** arguments are used allowing it to be actuated in order to enable the
```
jimis: Instead of "actuated" I would use the same term you used above, i.e. "evaluated", to avoid confusion. I would say something like:
> However this function has special behaviour when **exactly three** arguments are used, it is being evaluated even if it contains undefined variables. For example: ...

nickanderson:
thanks @jimis

nickanderson:
@jimis I can pick back to 3.7 and 3.9 if you can confirm the code is in those branches

jimis:
@nickanderson I just cherry-picked.

PR #1541, opened 2016-06-24
Change: Note behavior when exec_command missing

nickanderson:


reference/components/cf-execd.markdown
```diff
@@ -119,9 +119,11 @@ symbols may be used if desired.
 
 **Allowed input range:** `"?(/.*)`
 
+**Note:** If `exec_command` is not defined `cf-agent` will be executed with the `failsafe.cf` policy. If `$(sys.inputdir)/failsafe.cf` will be executed.
```
jimis: The phrase doesn't read right.

nickanderson: Ah yes. Ill redact the last partial sentence.

PR #1557, opened 2016-07-08
Add < <= > >= operators to eval().

tzz:
For https://github.com/cfengine/core/pull/2645 @jimis


jimis:
As always @tzz completing the job fully. Thanks!

PR #1559, opened 2016-07-13
Adapt docs according to fix for CFE-2412.

kacf:
Signed-off-by: Kristian Amlie <kristian.amlie@cfengine.com>


kacf:
Cherry-picked to 3.7 and 3.9 as well.

PR #1567, opened 2016-09-26
Fix history on if/unless in ifvarclass section

atsaloli:
According to ChangeLog in core, if/unless were introduced in 3.7.0,
not in 3.6.1:  https://github.com/cfengine/core/blob/master/ChangeLog#L759-L760


nickanderson:
Thank you!

PR #1569, opened 2016-09-30
CFE-2468 Document package module API

jpvlsmv:


jpvlsmv:
@nickanderson I'm not sure where the merge went wrong to shorten the first paragraph-delta (initially listing the out-of-the-box modules), but I think the longer documentation is the better of the two.

nickanderson:
I already merged it in master. Doc build is broken elsewhere. Working on it today.

jpvlsmv:
https://github.com/cfengine/documentation/blob/master/reference/promise-types/packages.markdown#yum still uses the apt_get module and "Apt::Install-Recommends=0" option, rather than something that would work with yum, like options => { "--disableplugin=fs-snapshot,allowdowngrade", "--skip-broken" };

(Cherrypick 1d8aea79694f7d44d15092b97e156672abef16b9 to see what I mean)

nickanderson:
@jpvlsmv thanks for pointing that out. Fixed in master docs: https://docs.cfengine.com/docs/master/reference-promise-types-packages.html#yum

PR #1571, opened 2016-10-06
ARCHIVE-1619: allow maplist(), maparray(), and mapdata() to evaluate function calls during iteration

tzz:
For https://github.com/cfengine/core/pull/2683#event-814928613


nickanderson:
thank you!

PR #1572, opened 2016-10-07
Add getuserinfo() function documentation.

tzz:
As requested in https://github.com/cfengine/core/pull/2575 @jimis @nickanderson

### DIFF
--- a/reference/functions/getuserinfo.markdown
+++ b/reference/functions/getuserinfo.markdown
@@ -1,0 +1,66 @@
+---
+layout: default
+title: getuserinfo
+published: true
+tags: [reference, user functions, functions, getuserinfo, users, uid, gid, gecos, homedir, shell]
+---
+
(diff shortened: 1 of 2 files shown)


reference/functions/getuserinfo.markdown
```diff
@@ -0,0 +1,64 @@
+---
+layout: default
+title: getuserinfo
+published: true
+tags: [reference, user functions, functions, getuserinfo, users, uid, gid, gecos, homedir, shell]
+---
+
+[%CFEngine_function_prototype(optional_uidorname)%]
+
+**Description:** Return information about the current user or any other, looked up by user ID (UID) or user name.
```
tzz: They are local in the sense that the system knows about them, so the function never looks in LDAP or anywhere else by itself. I rephrased it.

nickanderson:
Thank you!

PR #1574, opened 2016-10-07
Allow getvariablemetatags() and getclassmetatags() to get a specific tag key

tzz:
@nickanderson docs for https://github.com/cfengine/core/pull/2647


tzz:
These docs will automatically pick up the examples of the new usage from https://github.com/cfengine/core/issues/2717 when it's merged :)

PR #1589, opened 2016-12-29
Zendesk#3204: answer report collection questions.

jimis:
Also reduced header font size and simplified some preexisting info.


jimis:
Thank you Nick.

PR #1592, opened 2017-01-12
zendesk#3222: Improve pre-evalution documentation.

jimis:


jimis:
@nickanderson have a look

reference/language-concepts/normal-ordering.markdown
```diff
@@ -70,29 +70,29 @@ make sure all needed classes and variables are determined before they are
...
-is caused by the fact that both variables and classes placed in common bundles
+The following steps are executed per-bundle for each file parsed, in this order:
+1. if it's a common bundle, evaluate **vars** promises
+2. if it's a common bundle, evaluate **classes** promises
+3. evaluate **vars** promises
+(for details see `PolicyResolve()` in the C code)
+
+This is done because both variables and classes placed in common bundles
 are global whereas classes placed in agent bundles are local (by default) to
 bundles where those are defined. This means that during agent bundle
```
jimis: Is it true what is written here, regarding global and local vars/classes?

reference/language-concepts/normal-ordering.markdown
```diff
@@ -70,29 +70,29 @@ make sure all needed classes and variables are determined before they are
 needed in normal evaluation, pre-evaluation takes place immediately before
 policy evaluation.
 
-In pre-evaluation files are loaded based on ordering in body common control
+During pre-evaluation files are loaded based on ordering in body common control
 (first) and body file control (after body common control). This means that
 files included in body common control are loaded and parsed before files
 placed in body file control. This is important from a common bundles
```
nickanderson: Is the importance limited to *common* bundles? 

Ah yes it is. no need for change on line 76

reference/language-concepts/normal-ordering.markdown
nickanderson: For classes its true. Variables though don't really have local/global scope. However re-actuation of an agent bundle will clear some previously defined variables (classic arrays). I think I would just redact "both variables and" on line 86.

jimis: OK redacting

PR #1598, opened 2017-02-28
zendesk#3295: allowconnects, allowallconnects, trustkeysfrom

jimis:
* be clear on supported syntax
* remove hostnames from examples as they are not supported
* modify examples to use CIDR notation for subnets

### DIFF
--- a/reference/components/cf-serverd.markdown
+++ b/reference/components/cf-serverd.markdown
@@ -343,9 +343,9 @@ The full configuration would look something like this
 
         body server control
         {
-        allowconnects         => { "10.10.10" , "::1" };
-        allowallconnects      => { "10.10.10" , "::1" };
-        trustkeysfrom         => { "10.10.10" , "::1" };
+        allowconnects         => { "10.10.10.0/24" , "::1" };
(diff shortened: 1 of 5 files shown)


jimis:
@nickanderson please review and merge

reference/components/cf-serverd.markdown
```diff
@@ -67,27 +60,31 @@ will potentially match more than one hostname (e.g.,
...
+The clients that are not listed here may have only one open connection
+at the time with the server.
 
-See also the warning about regular expressions in
-[`allowallconnects`][cf-serverd#allowallconnects].
+Note that `127.0.0.1` is a regular expression (i.e., "127 any
+character 0 any character 0 any character 1"), but this will only
+match the IP address `127.0.0.1`. Take care with IP addresses and
+domain names, as the hostname regular expression `www.domain.com`
+will potentially match more than one hostname (e.g.,
```
mikeweilgart: I thought hostnames aren't accepted at all?

jimis: Yes you are right. This is a generic note for regexes (pre-existing, not added by me), but it should probably be under an attribute that accepts hostnames.

PR #1605, opened 2017-04-06
Add docs for functions string_head/tail negative arguments.

kacf:
Signed-off-by: Kristian Amlie <kristian.amlie@cfengine.com>

### DIFF
--- a/reference/functions/string_head.markdown
+++ b/reference/functions/string_head.markdown
@@ -9,7 +9,12 @@ tags: [reference, text functions, functions, text, head, string_head, substring]
 
 **Description:** Returns the first `max` bytes of `data`.
 
-[%CFEngine_function_attributes(data, max)%]
+If `max` is negative, then everything but the last `max` bytes is returned.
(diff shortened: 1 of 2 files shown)


reference/functions/string_head.markdown
```diff
@@ -9,7 +9,12 @@ tags: [reference, text functions, functions, text, head, string_head, substring]
 
 **Description:** Returns the first `max` bytes of `data`.
 
-[%CFEngine_function_attributes(data, max)%]
+If `max` is negative, then everything but the last `max` bytes is returned.
+
+**Arguments:**
+
+* `data`: `string`, in the range: `.*`
+* `max`: `int`, in the range: `-99999999999,99999999999`
```
kacf: @michaelclelland, @nickanderson: I couldn't find where the automatic attributes are extracted from, so I had to expand it like this in order to change from `0,99999999999` to `-99999999999,99999999999`. Do you know?

PR #1607, opened 2017-04-26
CFE-2513: New sysctlvalue() and data_sysctlvalues() functions from /proc/sys

tzz:
@kacf as requested in https://github.com/cfengine/core/pull/2743


kacf:
Thank you for the quick response! :-)

PR #1609, opened 2017-05-03
maxconnections

jimis:


jimis:
@nickanderson

nickanderson:
Should we mention that the processes open file handles are controlled by this value as of ...?

jimis:
These are internal details that might change if we see that the limit is hit again. I wouldn't document it but I can amend a small note if you insist.

nickanderson:
I think it would be nice to have it noted somewhere, perhaps just a small history entry like

**History:** 3.10.1, 3.11.0, 3.7.6 automatic open file limits set

PR #1611, opened 2017-05-15
Document and use optional regex parameter to --show-vars and --show-classes

tzz:
As discussed in https://github.com/cfengine/core/pull/2480


tzz:
@nickanderson can you add docs for the new `--show-evaluated-classes` and `--show-evaluated-vars` as well? I'm a bit underwater right now.

nickanderson:
Yes.

PR #1612, opened 2017-05-23
Added cf-net documentation

olehermanse:
Must be merged with https://github.com/cfengine/documentation-generator/pull/203

cf-net is 3.11+ so this is only for 3.11+.


olehermanse:
@nickanderson @kacf @jimis Please review.

reference/components/cf-net.markdown
```diff
@@ -0,0 +1,138 @@
...
+```
+$ sudo /var/cfengine/bin/cf-key
+$ sudo /var/cfengine/bin/cf-agent --bootstrap myhostname
+$ sudo /var/cfengine/bin/cf-net connect
+Connected & authenticated successfully to 'myhostname'
+```
+(`myhostname` can also be an IP address)
+
+All three commands above are run with sudo, so they access the same key file.
+In the future, we would like to make it easier to use cf-net without bootstrapping or sudo.
```
kacf: I would drop this "in the future..." sentence, statements such as these tend to linger for a long time if we don't end up doing it. I think it belongs in a ticket instead.

reference/components/cf-net.markdown
```diff
@@ -0,0 +1,138 @@
...
+
+## cf-net commands
+`cf-net` usage follows the general structure:
```
kacf: I would argue *syntax* instead of *usage*.

reference/components/cf-net.markdown
nickanderson: Agreed let's drop that sentence

reference/components/cf-net.markdown
nickanderson: I have no opinion here.

olehermanse:
Followed @kacf's suggestions and fixed both cases. Please merge

PR #1613, opened 2017-06-01
Add invenory api documentation

aleksandrychev:


reference/enterprise-api-ref/inventory.markdown
```diff
@@ -0,0 +1,213 @@
...
+        {
+          "columnName": "CFEngine version",
+          "columnType": "STRING"
+        },
+        {
+          "columnName": "Ports listening",
+          "columnType": "STRING"
+        }
+      ],
+      "query": "SELECT  values->>'Host name' AS \"Host name\",  values->>'OS type' AS \"OS type\",  cf_clearSlist(values->>'IPv4 addresses') AS \"IPv4 addresses\",  values->>'CFEngine version' AS \"CFEngine version\",  cf_clearSlist(values->>'Ports listening') AS \"Ports listening\" FROM inventory_new WHERE values->>'Host name' LIKE '%mender%' AND values->>'Host name' NOT ILIKE '%io%' ORDER BY values->>'Host name' ASC ",
```
maciejmrowiec: isn't this leaking internal implementation detail ?

reference/enterprise-api-ref/inventory.markdown
```diff
@@ -0,0 +1,213 @@
...
+      "queryTimeMs": 30,
+      "rowCount": 2,
+      "rows": [
+        [
+          host1.cfengine.com",
```
maciejmrowiec: missing start quote

reference/enterprise-api-ref/inventory.markdown
```diff
@@ -0,0 +1,213 @@
...
+          "linux",
+          "128.30.23.33",
+          "3.10.0",
+          "22, 25, 80, 443, 5308, 9000"
+        ],
```
maciejmrowiec: what if a host does not have a specific column, how would that response look? (maybe include in example )

reference/enterprise-api-ref/inventory.markdown
```diff
@@ -0,0 +1,213 @@
...
+      "query": "SELECT  values->>'Host name' AS \"Host name\",  values->>'OS type' AS \"OS type\",  cf_clearSlist(values->>'IPv4 addresses') AS \"IPv4 addresses\",  values->>'CFEngine version' AS \"CFEngine version\",  cf_clearSlist(values->>'Ports listening') AS \"Ports listening\" FROM inventory_new WHERE values->>'Host name' LIKE '%mender%' AND values->>'Host name' NOT ILIKE '%io%' ORDER BY values->>'Host name' ASC ",
+      "queryTimeMs": 30,
+      "rowCount": 2,
+      "rows": [
+        [
+          host1.cfengine.com",
+          "linux",
+          "128.30.23.33",
+          "3.10.0",
+          "22, 25, 80, 443, 5308, 9000"
```
maciejmrowiec: thats string or array ? (column type says string but looks like an array to me)

reference/enterprise-api-ref/inventory.markdown
aleksandrychev: You're right that is unnecessary information.

reference/enterprise-api-ref/inventory.markdown
aleksandrychev: ok

reference/enterprise-api-ref/inventory.markdown
aleksandrychev: it will be null, I am going to include it

reference/enterprise-api-ref/inventory.markdown
aleksandrychev: It is string starts and ends with quote

reference/enterprise-api-ref/inventory.markdown
```diff
@@ -0,0 +1,211 @@
...
+  "id": 1,
+  "attribute_name": "Architecture",
+  "category": "Hardware",
+  "readonly": 0,
+  "type": "slist",
+  "convert_function": "cf_clearSlist"
+}
+```
+
+## Add custom variables to variables dictionary
```
nickanderson: I don't understand what this is for. Is this to create a new category like Hardware or Software but user defined? Or is it for mapping a variable into an existing group?

aleksandrychev:
Curl examples were added.

aleksandrychev:
> Looks better, can you update the URIs used for the curl examples to be consistent https://hub.cfengine.com

Done.

reference/enterprise-api-ref/inventory.markdown
```diff
@@ -0,0 +1,255 @@
...
+    "page": 1,
+    "timestamp": 1496222472,
+    "total": 1
+  }
+}
+```
+
+## List of variables
+
+**URI:** https://hub.cfengine.com/api/variables-dictionary
```
nickanderson: @maciejmrowiec @michaelclelland is this a good API endpoint? We report on variables that are not part of inventory, this would be a bit misleading to me. By looking only at the api path I would assume this is where I would query on arbitrary variables not limited to those tagged for inventory.

reference/enterprise-api-ref/inventory.markdown
```diff
@@ -0,0 +1,255 @@
...
+```
+
+## Edit variable
+
+**URI:** https://hub.cfengine.com/api/variables-dictionary/:id
+
+**Method:** PATCH
+
+Set variables type (int/string..). This is needed for applying filtering in correct format.
+Only `readonly - 0` variable can be edited
```
nickanderson: I think variable should be plural. "Only `readonly - 0` variables can be edited.

nickanderson: and variable word itself may need replaced based on @michaelclelland and @maciejmrowiec comments from above.

nickanderson: Along the same lines as above with naming. Editing variable makes me thing we are editing variable values. Edit inventory attribute was first thought to come to mind, but I dont like the word attribute here. Maybe "Edit inventory trait"?

reference/enterprise-api-ref/inventory.markdown
```diff
@@ -0,0 +1,255 @@
...
+| lt           | <         |
+| gt           | >         |
+| eq           | =         |
+| neq          | !=        |
+| lteq         | <=        |
+| gteq         | >=        |
+| l            | LIKE      |
+| nl           | NOT LIKE  |
+| il           | ILIKE     |
+| nil          | NOT ILIKE |
```
nickanderson: Are these standard abbreviations?
I definitely would not have figured out what `l`, `nl`, `il`, and `nil` were related to database LIKE operators. I am also no API guru.

reference/enterprise-api-ref/inventory.markdown
```diff
@@ -0,0 +1,255 @@
...
+
+**Example Request Body:**
+
+
+```
+{  
+   "sort":"Host name",
+   "filter":{  
+      "Host name":{  
+         "l":"%value1%",
```
nickanderson: It took me some fiddling to understand what the % here meant.

Was it required on both ends as shown, is it automatically inserted around value. Maybe thats just my lack of experience around API docs.

I did get it working:
If % isnt used then it must be an exact match for hostname.
it can be used for wildcard on front or back or both.

aleksandrychev: @nickanderson  I have intent to change operators, it will probably 
'<', '>','!=','=','<=', '>=', 'match', 'not_match', 'contains', 'not_contains'.
it gives possibility to avoid using %, back-end will handle this cases by himself.

reference/enterprise-api-ref/inventory.markdown
aleksandrychev: There are no standards to do abbreviations with LIKE. I will change it. read the comment https://github.com/cfengine/documentation/pull/1613#discussion_r140219685

reference/enterprise-api-ref/inventory.markdown
nickanderson: cool

reference/enterprise-api-ref/inventory.markdown
aleksandrychev: As far as I understand I have to change endpoint to https://hub.cfengine.com/api/inventory/variables-dictionary ?

nickanderson: I believe it should be called attributes-dictionary

nickanderson: Still looks like the docs reference variables-dictionary instead if the new name

nickanderson:
Can you please squash the commits into 1 commit please

nickanderson:
Once squashed it can be Merged

aleksandrychev:
>Once squashed it can be Merged

Done

PR #1618, opened 2017-06-21
Ldap APi documentation was added.

aleksandrychev:
@nickanderson @dmitry84 Plese review

### DIFF
--- a/reference/enterprise-api-ref/ldap-api.markdown
+++ b/reference/enterprise-api-ref/ldap-api.markdown
@@ -1,0 +1,125 @@
+---
+layout: default
+title: LDAP authentication API
+published: true
+tags: [reference, enterprise, API, authenticating, LDAP]
+---
+
+LDAP authentication API allows to check ldap user credentials and change LDAP settings.
+
+## Login
+
+**URI:** https://hub.cfengine.com/ldap/login
+
+**Method:** POST
+
+**Parameters:**
+
+* **username** *(string)*
+    Username from LDAP
+* **password** *(string)*
+    User password
+
+**Example response:**
+
+```
+HTTP 200 Ok
+{
+    "success": true,
+    "message": "You are successfully authenticated"
+}
+```
+
+## Get settings
+
+**URI:** https://hub.cfengine.com/ldap/settings
+
+**Method:** GET
+
+**Headers:**
+
+* **Authorization: api_token** *(string)*
+    Set token to access api
+
+**Example response:**
+
+```
+HTTP 200 Ok
+{
+    "success": true,
+    "data": {
+        "domain_controllers": "local.loc",
+        "base_dn": "DC=local,DC=loc",
+        "login_attribute": "cn",
+        "port": 365,
+        "use_ssl": false,
+        "use_tls": false,
+        "timeout": 5,
+        "admin_username": "cn=admin,DC=local,DC=loc",
+        "admin_password": "Password is set"
+    }
+}
+```
+
+**Output:**
+
+* **domain_controllers**
+    The domain controllers option is   server name located on your network that serve Active Directory.
+* **base_dn**
+    The base distinguished name is the base distinguished name you'd like to perform operations on. An example base DN would be DC=corp,DC=acme,DC=org.
+* **login_attribute**
+    Login attribute like cn or uid
+* **port**
+    The port option is used for authenticating and binding to your AD server. The default ports are already used for non SSL and SSL connections (389 and 636).
+* **use_ssl**
+    User ssl for connection
+* **use_tls**
+    User tls for connection
+* **timeout**
+    The timeout option allows you to configure the amount of seconds to wait until your application receives a response from your LDAP server.
+* **admin_username**
+    LDAP admin distinguished name
+* **admin_password**
+    LDAP admin password.
+
+## Update settings
+
+**URI:** https://hub.cfengine.com/ldap/settings
+
+**Method:** PATCH
+
+**Headers:**
+
+* **Authorization: api_token** *(string)*
+    Set token to access api
+
+**Parameters:**
+
+* **domain_controllers** *(string)*
+    The domain controllers option is   server name located on your network that serve Active Directory.
+* **base_dn** *(string)*
+    The base distinguished name is the base distinguished name you'd like to perform operations on. An example base DN would be DC=corp,DC=acme,DC=org.
+* **login_attribute**  *(string)*
+    Login attribute like cn or uid
+* **port** *(integer)*
+    The port option is used for authenticating and binding to your AD server. The default ports are already used for non SSL and SSL connections (389 and 636).  Optional parameter.
+* **use_ssl** *(boolean)*
+    User ssl for connection.  Optional parameter.
+* **use_tls** *(boolean)*
+    User tls for connection.  Optional parameter.
+* **timeout** *(integer)*
+    The timeout option allows you to configure the amount of seconds to wait until your application receives a response from your LDAP server.  Optional parameter.
+* **admin_username**
+        LDAP admin distinguish name
+* **admin_password**
+        LDAP admin password.
+
+**Example response:**
+
+```
+HTTP 200 Ok
+{
+    "success": true,
+    "message": "Settings successfully saved."
+}
+```
--- a/reference/enterprise-api-ref/ldap-api.markdown
+++ b/reference/enterprise-api-ref/ldap-api.markdown
@@ -39,7 +39,7 @@ HTTP 200 Ok
 **Headers:**
 
 * **Authorization: api_token** *(string)*
-    Set token to access api
+    Set token to access api. To get the token please look at - ```/var/cfengine/httpd/htdocs/ldap/config/settings.php```
 
 **Example response:**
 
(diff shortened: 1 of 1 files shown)


reference/enterprise-api-ref/ldap-api.markdown
```diff
@@ -0,0 +1,125 @@
...
+    "data": {
+        "domain_controllers": "local.loc",
+        "base_dn": "DC=local,DC=loc",
+        "login_attribute": "cn",
+        "port": 365,
+        "use_ssl": false,
+        "use_tls": false,
+        "timeout": 5,
+        "admin_username": "cn=admin,DC=local,DC=loc",
+        "admin_password": "Password is set"
```
nickanderson: Literally it returns `Password is set`?

This differs from pre-ldap api settings correct?

Before the new ldap api ldap password was just stripped from the response correct?

reference/enterprise-api-ref/ldap-api.markdown
```diff
@@ -0,0 +1,125 @@
...
+    }
+}
+```
+
+**Output:**
+
+* **domain_controllers**
+    The domain controllers option is   server name located on your network that serve Active Directory.
```
nickanderson: Are we allowed to have multiple domain controllers? The name makes me think there can be more than one. Is this stubbed in for future support?

reference/enterprise-api-ref/ldap-api.markdown
```diff
@@ -0,0 +1,125 @@
...
+* **base_dn**
+    The base distinguished name is the base distinguished name you'd like to perform operations on. An example base DN would be DC=corp,DC=acme,DC=org.
+* **login_attribute**
+    Login attribute like cn or uid
+* **port**
+    The port option is used for authenticating and binding to your AD server. The default ports are already used for non SSL and SSL connections (389 and 636).
+* **use_ssl**
+    User ssl for connection
```
nickanderson: s/User/Use

reference/enterprise-api-ref/ldap-api.markdown
```diff
@@ -0,0 +1,125 @@
...
+* **use_tls**
+    User tls for connection
```
nickanderson: s/User/Use/

reference/enterprise-api-ref/ldap-api.markdown
```diff
@@ -0,0 +1,125 @@
...
+* **timeout**
+    The timeout option allows you to configure the amount of seconds to wait until your application receives a response from your LDAP server.
+* **admin_username**
```
nickanderson: example please

reference/enterprise-api-ref/ldap-api.markdown
```diff
@@ -0,0 +1,125 @@
...
+## Update settings
+
+**URI:** https://hub.cfengine.com/ldap/settings
+
+**Method:** PATCH
+
+**Headers:**
+
+* **Authorization: api_token** *(string)*
+    Set token to access api
```
nickanderson: How does one get this API token?

reference/enterprise-api-ref/ldap-api.markdown
```diff
@@ -0,0 +1,125 @@
...
+
+**Parameters:**
+
+* **domain_controllers** *(string)*
```
nickanderson: Multiple?

reference/enterprise-api-ref/ldap-api.markdown
```diff
@@ -0,0 +1,125 @@
...
+    The domain controllers option is   server name located on your network that serve Active Directory.
+* **base_dn** *(string)*
+    The base distinguished name is the base distinguished name you'd like to perform operations on. An example base DN would be DC=corp,DC=acme,DC=org.
+* **login_attribute**  *(string)*
+    Login attribute like cn or uid
+* **port** *(integer)*
+    The port option is used for authenticating and binding to your AD server. The default ports are already used for non SSL and SSL connections (389 and 636).  Optional parameter.
+* **use_ssl** *(boolean)*
+    User ssl for connection.  Optional parameter.
```
nickanderson: s/User/Use/

reference/enterprise-api-ref/ldap-api.markdown
```diff
@@ -0,0 +1,125 @@
...
+* **use_tls** *(boolean)*
+    User tls for connection.  Optional parameter.
```
nickanderson: s/User/Use/

reference/enterprise-api-ref/ldap-api.markdown
```diff
@@ -0,0 +1,125 @@
...
+* **timeout** *(integer)*
+    The timeout option allows you to configure the amount of seconds to wait until your application receives a response from your LDAP server.  Optional parameter.
+* **admin_username**
+        LDAP admin distinguish name
```
nickanderson: example please

reference/enterprise-api-ref/ldap-api.markdown
aleksandrychev: It returns "Password is set" to avoid showing the password.
>This differs from pre-ldap api settings correct? Before the new ldap api ldap password was just stripped from the response correct?

Yes, before the password was stripped from the response. I can remove password field from the response.

reference/enterprise-api-ref/ldap-api.markdown
aleksandrychev: New API gives us the possibility to set more than one domain controller, but current UI implementation isn't. So I cant rename domain_controllers to domain_controller.

reference/enterprise-api-ref/ldap-api.markdown
aleksandrychev: done

reference/enterprise-api-ref/ldap-api.markdown
aleksandrychev: done

reference/enterprise-api-ref/ldap-api.markdown
aleksandrychev: What do you mean? 
>"timeout": 5,   "admin_username": "cn=admin,DC=local,DC=loc",

reference/enterprise-api-ref/ldap-api.markdown
aleksandrychev: https://github.com/cfengine/documentation/pull/1618#discussion_r143965118

reference/enterprise-api-ref/ldap-api.markdown
aleksandrychev: done

reference/enterprise-api-ref/ldap-api.markdown
aleksandrychev: done

reference/enterprise-api-ref/ldap-api.markdown
aleksandrychev: added

reference/enterprise-api-ref/ldap-api.markdown
aleksandrychev: API token i /ldap/config/settings.php https://github.com/cfengine/ldap-api/blob/master/config/settings.php#L17 .

aleksandrychev: I am not sure about posting this documentation as public... Because it's more internal stuff. And using as separate part makes no sense. I this we will discuss it.

reference/enterprise-api-ref/ldap-api.markdown
nickanderson: I don't know what is more correct or desirable just asking for clarification

reference/enterprise-api-ref/ldap-api.markdown
nickanderson: If the field can have multiple then we can leave it. Ui Support isn't important for the API imho

aleksandrychev: Also, I  I've implemented saving the setting only with one domain, so, or we should change the behavior of saving, or rename this attribute.

dmitry84: I think it is better to rename to domain_controller
because even if we can save multiple domains via LDAP library API, our API and UI  won't be able to show and save this.

reference/enterprise-api-ref/ldap-api.markdown
aleksandrychev: I've added info about token.

reference/enterprise-api-ref/ldap-api.markdown
```diff
@@ -0,0 +1,125 @@
...
+* **domain_controller**
+    The domain controllers option is   server name located on your network that serve Active Directory.
+* **base_dn**
+    The base distinguished name is the base distinguished name you'd like to perform operations on. An example base DN would be DC=corp,DC=acme,DC=org.
+* **login_attribute**
+    Login attribute like cn or uid
+* **port**
+    The port option is used for authenticating and binding to your AD server. The default ports are already used for non SSL and SSL connections (389 and 636).
+* **use_ssl**
+    User ssl for connection
```
nickanderson: I still see "User ssl for connection" here.

reference/enterprise-api-ref/ldap-api.markdown
nickanderson: I still see "User tls for connection" here instead of "Use tls for connection"

PR #1651, opened 2017-11-13
zd#3599: simplify upgrade process documentation

jimis:


jimis:
@nickanderson please review. My goal is to have a short summary on top that shows the whole process is not **that** cumbersome, before the section with all the excruciating details.

jimis:
Merging since it's a simple change.

PR #1668, opened 2017-11-28
The timeout parameter is actually a connection + read timeout

amousset:


nickanderson:
Thanks!

Backported to 3.7 and 3.10

PR #1683, opened 2017-12-08
ENT-3486 Host count API documentation.

aleksandrychev:


nickanderson:
@aleksandrychev  I think this looks OK. but I am curious about the date format for the api. I think in many other places we are using unix epoch time. What formats are valid input to this api and is there any way to specify the time spec for output (not that we necessarily need to have that ability)? 

@estenberg  @craig.comstock  @michaelclelland  have you any thoughts about time format and useability here?

nickanderson:
For pondering and consideration. 
http://apiux.com/2013/03/20/5-laws-api-dates-and-times/

aleksandrychev:
>For pondering and consideration. http://apiux.com/2013/03/20/5-laws-api-dates-and-times/
I don't think we need date and time, in this api we use only date ranges.

nickanderson:
@aleksandrychev is this ready to be merged (is the API already merged?)

PR #1711, opened 2018-01-15
Adapt to the changes in the mapdata() example

vpodzime:
The result of mapdata() is not deterministic. So the example
needs to use a trick to show some deterministic example data
instead.


vpodzime:
@nickanderson please review

PR #1734, opened 2018-02-20
CFE-2310: Document the IPv6 network hard classes

vpodzime:
Together with an example of how they look like.


nickanderson:
Working backwards :-P

nickanderson:
Thank you!

PR #1735, opened 2018-02-22
filesexist() is now one of the collecting functions

vpodzime:
Implemented in cfengine/core#3033


vpodzime:
> Can you re-write the commit message so that the JIRA issue id starts the subject?

Done.

PR #1742, opened 2018-03-09
Update dead link for Enterprise Guide Reporting

mayoralade:


Lex-2008:
To make it more future-proof, on master branch I'd change all links on this page to relative ones.

Note: this file is same on master and 3.10 branches, so this commit can be safely backported.

mayoralade:
@Lex-2008 I updated the links to relative ones.

examples/tutorials/tags.markdown
```diff
@@ -48,7 +48,7 @@ bundle agent cfe_autorun_inventory_listening_ports
 
 In the Enterprise Mission Portal, you can then make a report for
 "Ports listening" across all your machines. For more details, see
-https://cfengine.com/docs/master/guide-enterprise-cfengine-guide-mission-portal-reporting.html
+[Enterprise Reporting](../../enterprise-cfengine-guide/reporting.markdown)
```
nickanderson: Here instead of providing an explicit path, we should be able to use the page title as the target.

[Enterprise Reporting][Enterprise Reporting].

The text between the first brackets are the anchor text, and the text between the second brackets is the page title as specified in the page front matter of the page the link should target.

https://raw.githubusercontent.com/cfengine/documentation/master/enterprise-cfengine-guide/reporting.markdown

nickanderson:
Thanks for this work @mayoralade.  I incorporated it as part of a refresh https://github.com/cfengine/documentation/pull/1743

PR #1762, opened 2018-03-23
ENT-3695 Refresh secure bootstrap guide

nickanderson:
Refresh the guide to include the new --trust-key option.

### DIFF
--- a/guide/installation-and-configuration/secure-bootstrap.markdown
+++ b/guide/installation-and-configuration/secure-bootstrap.markdown
@@ -6,7 +6,6 @@ sorting: 20
 tags: [guide, installation, install, security]
 ---
 
-
 This guide presumes that you already have CFEngine properly installed
 and running on the policy hub, the machine that distributes the policy
 to all the clients. It also presumes that CFEngine is installed, but not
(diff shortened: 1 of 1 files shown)


guide/installation-and-configuration/secure-bootstrap.markdown
```diff
@@ -37,125 +35,87 @@ already have a secure channel to the clients, usually ssh, and we use
...
 settings. So after we have set-up our hub (using the standard procedure
-of ```cf-agent --bootstrap $HUB_IP```) we take care of the following:
+of `cf-agent --bootstrap $HUB_IP`) we take care of the following:
 
-* Cf-serverd must never accept a connection from a client presenting an
-  untrusted key.  So in `body server control` we must empty the
-  `trustkeysfrom` attribute:
+* `cf-serverd` must never accept a connection from a client presenting an
+  untrusted key. [Disable automatic key trust][Masterfiles Policy
+  Framework#trustkeysfrom] by providing an empty list for `def.trustkeyfrom`.
```
vpodzime: The second pair of brackets should be parentheses, I think.

guide/installation-and-configuration/secure-bootstrap.markdown
```diff
@@ -37,125 +35,87 @@ already have a secure channel to the clients, usually ssh, and we use
...
-  ```
-  cp /var/cfengine/inputs/failsafe.cf /var/cfengine/masterfiles/
-  ```
+```console
+[root@host001]# cf-key --trust-key /path/to/hubs/key.pub
+```
 
-  We'll edit that copy in `masterfiles` in the next step.
+**Note:** If you are using [protocol_version `1` or `classic`][Components and Common Control#protocol_version]
+you need to supply an IP address before the path to the key.
```
vpodzime: Same here.

guide/installation-and-configuration/secure-bootstrap.markdown
vpodzime: Oh, okay, good to know.

PR #1801, opened 2018-05-09
ENT-3550: Change preference documentation.

aleksandrychev:


aleksandrychev:
@nickanderson please review

nickanderson:
This looks OK to me. But I have a question about API versioning that I posted into ENT-3550 and pinged the other devs for perspective. After we hear back on that this can move forward

craigcomstock:
I'm inclined to not version it since it is an addition and no change/removal of existing API. FWIW one source to back me up. Not saying this source is authoritative: https://restfulapi.net/versioning/

nickanderson:
@aleksandrychev  do we need to merge anything along with this?
 Or is this ready to go in now?It will need to be picked back to the 3.12 branch as well.

aleksandrychev:
>do we need to merge anything along with this?

yes, https://github.com/cfengine/nova/pull/1169 and https://github.com/cfengine/mission-portal/pull/542

>It will need to be picked back to the 3.12 branch as well.

OK

PR #1856, opened 2018-06-30
CFE-2837 Improve docs for yum enable/disable repo options

nickanderson:


atsaloli:
I checked it on https://docs.cfengine.com/docs/3.12/reference-promise-types-packages.html and it looks good, thanks!

PR #1863, opened 2018-07-11
ENT-3945 Update supported platforms for 3.12.x

nickanderson:
Support for the following platform versions was removed because they have not been supported by their vendor for at least one year.

- AIX 5.3, 6 (PowerPC)
- CentOS/RHEL 4 (x86-64, x86)
- HP-UX 11.23 (Itanium)
- SLES 10 (x86-64, x86)
- Solaris 9 (UltraSparc)
- Ubuntu 10.04, 12.04 (x86-64, x86)

### DIFF
--- a/guide/latest-release/supported-platforms.markdown
+++ b/guide/latest-release/supported-platforms.markdown
@@ -22,18 +22,18 @@ Any supported host can be a policy server in Community installations of CFEngine
 
 ## Hosts ##
 
-| Platform    | Versions                   | Architectures   |
-| :-----:     | :----------:               | :-----------:   |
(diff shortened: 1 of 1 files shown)


roschern:
AIX 5.3 and Solaris 9 have already been removed before the release of 3.12, so they do not need to be mentioned.

roschern:
We also need to add the specific policy wording to https://cfengine.com/product/supported-versions/
Should that that be a separate PR?

nickanderson:
https://cfengine.com/product/supported-versions/ is a completely different website. You edit that one directly, no PR.

roschern:
then this looks good to me.

PR #1881, opened 2018-07-18
CFE-2850: Fixed wrong explanation of trivial option in regex_replace

karlhto:
The documentation previously claimed that the trivial option `T'
disabled searching for regex patterns, only searching for exact
strings. What it is supposed to do, as per what PCRE defines, is
remove backreferences and special characters in the replacement
string.

Ticket: CFE-2850
Changelog: Title


karlhto:
@nickanderson Done.

nickanderson:
Hum, I still see that you are "unverified" (no gpg signature on the commit).

nickanderson:
Ah, I think you need to upload your gpg key to github.

karlhto:
@nickanderson You were right. The key I had added was apparently not updated.

nickanderson:
Thanks!

PR #1887, opened 2018-07-23
Minor augments docs improvements

vpodzime:


cf-bottom:

nickanderson:
@craigcomstock 
Maybe for version 4.0? :-P it's been the way it is for so long I think it would probably be bad to change it during 3.x

nickanderson:
thanks @vpodzime !

craigcomstock:
Yes @nickanderson I agree of course. It might be hard to change and keep backward compatibility. ;/

PR #1898, opened 2018-07-25
Cross reference max_children and background attributes

nickanderson:


craigcomstock:
thanks @nickanderson !

PR #1952, opened 2018-10-08
CFE-2915 Align example in docs with bodies that exist in the stdlib

nickanderson:
While the example is technically correct, using the body name from the stdlib
will encourage use of existing bodies instead of duplicating them as someone
trying to copy/paste will receive a duplicate definition of body error.

(cherry picked from commit fa0aecf34963189adfeefc9860d80fb1ee7d5278)


cf-bottom:

PR #1969, opened 2018-10-23
CFE-2929 Note that getindices() always returns a list

nickanderson:
As indicated in CFE-1893 and core commit f235637ccfa756d91f042f5254f44631a8840389, the function will always return a list.


cf-bottom:

PR #1974, opened 2018-10-30
small typo in package module docs

craigcomstock:


nickanderson:
I tried to pick it to 3.12.x and open a PR but i reached my api rate limit.

nickanderson:
I went ahead and pushed directly to 3.12, 3.10. and 3.7 so this is all done. no need to mess with it.

PR #1987, opened 2018-11-07
fix location of report_data_select attributes 3.10

nickanderson:


cf-bottom:

PR #1992, opened 2018-11-13
Remove an old exception in hard classes usable in augments

vpodzime:
cfengine/core@5214d9347 changed the order in which things are
evaluated so now even the `am_policy_hub` and `policy_server`
hard classes can be used in augments.

Ticket: CFE-2941


cf-bottom:

PR #1994, opened 2018-11-14
CFE-2853 Note that cf-execd and cf-serverd respond to SIGHUP

nickanderson:
This behavior was added in 3.7.0 as part of CFE-1698


cf-bottom:

PR #1996, opened 2018-11-14
CFE-2853 Note that cf-execd and cf-serverd respond to SIGHUP 3.10.x

nickanderson:
This behavior was added in 3.7.0 as part of CFE-1698

(cherry picked from commit a9482bd34e12c1b15c0da04f43c9094786b3a0ea)


atsaloli:
:+1:

PR #2001, opened 2018-11-16
Fixed order of systemctl/service args

olehermanse:


cf-bottom:

PR #2003, opened 2018-11-16
Fix note about regular expression for bundlesmatching being unanchored

nickanderson:
It's really anchored.


nickanderson:
Thanks, @mikeweilgart, and @atsaloli

PR #2036, opened 2018-12-05
faq-slow-query-log: Instructions for enabling slow query log for postgresql 3.7

nickanderson:


cf-bottom:

PR #2037, opened 2018-12-12
CFE-2954 Clarify how classes are defined from augments

nickanderson:
Especially with regard to defining classes based on classes not being defined
yet using negative lookahead.


cf-bottom:

PR #2038, opened 2018-12-12
CFE-2954 Clarify how classes are defined from augments

nickanderson:
Especially with regard to defining classes based on regular expressions that
look for no class to match.

Changelog: None
(cherry picked from commit bbca56eb04594024aa61fa21b0d32cd8b82e243c)


cf-bottom:

PR #2042, opened 2018-12-14
ENT-3294 Document monitoring tables

nickanderson:
Changelog: None


nickanderson:
See also https://github.com/cfengine/nova/pull/1324

PR #2043, opened 2018-12-18
ENT-3633 Add note about windows services not using display name

nickanderson:


cf-bottom:

PR #2049, opened 2018-12-18
CFE-2478: Fixed documentation entry for `module`

karlhto:
Since forward slash will be added in
https://github.com/cfengine/core/pull/3472, I have altered the
documentation entry to no longer point to CFE-2478, and added
that forward slash is allowed.

Ticket: CFE-2478
Changelog: None

Blocked by cfengine/core#3472


cf-bottom:
Thank you for submitting a PR! Maybe @nickanderson can review this?

PR #2059, opened 2019-01-03
ENT-4149: Change reset administrative credentials documentation

aleksandrychev:
Do not overwrite email and roles in case if admin user exists


aleksandrychev:
@craigcomstock @nickanderson please review

PR #2064, opened 2019-01-10
Added small clarifications about ordering of vars promises

olehermanse:


reference/promise-types/vars.markdown
```diff
@@ -376,3 +376,5 @@ two_example_com::
      string => "two.example.com",
     comment => "Define a global domain for hosts in the two.example.com domain";
 ```
+
+(Promises within the same bundle are normally evaluated top to bottom, so vars promises further down in a bundle can overwrite previous values of a variable. See [**normal ordering**][Normal Ordering] for more information).
```
nickanderson: I would redact 'normally'. They are always evaluated top to bottom for a given promise type within a bundle. Order of actuation of each promise can be controlled wihtin a bundle using depends_on or classes to delay actuation until a subsequent pass (there are 3 passes within a bundle).

Otherwise, looks great

olehermanse: Fixed!

nickanderson:
picked to 3.12, 3.10 and 3.7

PR #2073, opened 2019-02-18
ENT-4445: Document how hosts are deleted

olehermanse:


olehermanse:
@craigcomstock @nickanderson @aleksandrychev please review.

nickanderson:
@olehermanse Looks good. I will make a few minor modifications myself after merging.

This can be picked to 3.12.x, but since 3.10 includes redis, I guess not back to 3.10.x?

olehermanse:
@nickanderson Actually, I think this is the same in 3.10. Only cross check needed is for the response codes, introduced recently by @aleksandrychev.

nickanderson:
Picked to 3.12.x

nickanderson:
@aleksandrychev are the return codes documented in this PR accurate for 3.10.x?

nickanderson:
@aleksandrychev Never mind, it is: https://github.com/cfengine/nova/pull/1361

PR #2080, opened 2019-02-18
ENT-4462: Clarify why host count in cf-hub and Mission Portal can differ

olehermanse:


olehermanse:
@nickanderson please review.

PR #2093, opened 2019-04-05
Fix tip in lsdir documentation

mikeweilgart:
The previous tip would actually omit all files whose names start with a period, not just "." and ".."


cf-bottom:

nickanderson:
Very nice, thank you!

PR #2094, opened 2019-04-08
$(sys.policy_hub) is the IP address of the policy hub

vpodzime:
Not the hostname.


cf-bottom:

nickanderson:
So, I believe that technically, whatever you bootstrap to is stored in policy_server.dat. In order to maintain backward compatibility, we resolve hostnames to the current IP and always present sys.policy_server as an IP address. So, bootstrapping to a hostname allows the agent to follow changes with DNS records.

nickanderson:
superseded by  https://github.com/cfengine/documentation/pull/2097

PR #2110, opened 2019-04-23
CFE-3040: Fix process promise example (3.12)

nickanderson:
process_owner is not an attribute, it's part of the process_select body.

Ticket: CFE-3040
Changelog: None


cf-bottom:

PR #2126, opened 2019-05-24
ENT-4299 Reorder and add missing SQL schema documentation (3.12)

aleksandrychev:
ChangeLog: Title


aleksandrychev:
@nickanderson I re-checked SQL schema in 3.12 and master, tables which are accessible via QUERY API are equal, so I just copied master's documentation. there are reordered tables by name. Please review and merge if it's OK

PR #2168, opened 2019-07-12
Switch eval() to use example and output from core/examples

nickanderson:


cf-bottom:

PR #2171, opened 2019-07-12
CFE-3099/3.12 Note that @ is allowed in variables defined by module protocol.

nickanderson:


cf-bottom:

PR #2174, opened 2019-07-12
CFE-3101/3.12 Fixed note about iterating over global lists and use core example

nickanderson:


cf-bottom:

PR #2177, opened 2019-07-15
Add some details about the hostrange() function

vpodzime:
The documentation of the behavior and formats of arguments needs
to be as precise as possible.

Ticket: CFE-3104


cf-bottom:

PR #2184, opened 2019-07-26
CFE-3103/3.10: Clarified output captured by execresult() includes both stdout and stderr

nickanderson:


cf-bottom:

PR #2189, opened 2019-08-19
Add note to get around python3 being hard to install on centos6

craigcomstock:


craigcomstock:
Just tested this again and confirmed the problem and this solution as working.

PR #2203, opened 2019-10-05
CFE-3171/master: Added references to dotall and multiline modes for data_regextract()

nickanderson:
Merge together:
https://github.com/cfengine/core/pull/3871


mikeweilgart:
Thanks, Nick.  Question for you - I gave you that link for the pcre2 regex syntax summary, but...is CFEngine actually using pcre2, or the older pcre?  If it's not using pcre2 the link should be fixed to avoid confusion.

PR #2207, opened 2019-10-08
CFE-3165/3.12: Switch to using example from core to illustrate function return types

nickanderson:


nickanderson:
The example fails in 3.12, so recinding this.

PR #2226, opened 2019-10-18
Fix link to point to the version of pcre we actually use

nickanderson:


nickanderson:
@mikeweilgart I remember seeing a message from you about this but i couldn't find it to respond.

Thanks, indeed we don't use pcre2, we use the original pcre 8.43 https://github.com/cfengine/buildscripts/tree/master/deps-packaging#agent-dependencies

mikeweilgart:
> @mikeweilgart I remember seeing a message from you about this but i couldn't find it to respond.
> 
> Thanks, indeed we don't use pcre2, we use the original pcre 8.43 https://github.com/cfengine/buildscripts/tree/master/deps-packaging#agent-dependencies

Thanks!  It was here: https://github.com/cfengine/documentation/pull/2203#issuecomment-542459341

PR #2233, opened 2019-11-15
CFE-3194: Documentation for negative indices in Nth

karlhto:
Ticket: CFE-3194
Changelog: None

### DIFF
--- a/reference/functions/nth.markdown
+++ b/reference/functions/nth.markdown
@@ -9,9 +9,8 @@ tags: [reference, data functions, functions, nth, inline_json]
 
 **Description:** Returns the element of `list_or_container` at zero-based `position_or_key`.
 
-If an invalid position (below 0 or above the size of the list minus 1)
-or missing key is requested, this function does not return a valid
-value.
+If an invalid position (above the size of the list minus 1) or missing key is
+requested, this function does not return a valid value.
 
 [This function can accept many types of data parameters.][Functions#collecting functions]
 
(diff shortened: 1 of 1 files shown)


cf-bottom:

reference/functions/nth.markdown
```diff
@@ -9,9 +9,8 @@ tags: [reference, data functions, functions, nth, inline_json]
 
 **Description:** Returns the element of `list_or_container` at zero-based `position_or_key`.
 
-If an invalid position (below 0 or above the size of the list minus 1)
-or missing key is requested, this function does not return a valid
-value.
+If an invalid position (above the size of the list minus 1) or missing key is
+requested, this function does not return a valid value.
```
nickanderson: a valid value?
So, if I do "string" string => nth( list, -100 )where list has only 10 elements, what will string be set to?

karlhto: It won't return a valid value. Could clarify by putting "or less than the negative of the length" or similar? Not a very good phrasing, so if you have a suggestion for how it could be phrased better, feel free to add it

nickanderson: Does it error then?

karlhto: No, but I agree that it should, although changing it might break some setups. An out of bound index is just ignored by the function (as it is written now). Our acceptance test also has out of bound values that it tests against, where the function returning nothing means success.

nickanderson: Maybe a snippet that shows the effect of providing an invalid position would help. Can you link me to a build where this is implemented?

karlhto: Here's the PR for the change, where you have changes to the acceptance test as well: https://github.com/cfengine/core/pull/3944

The build was triggered 6 days ago, so I don't think it's reachable anymore. It was merged to master today so nightly builds aren't usable before tomorrow I think.

```
"test" slist => { 1, 2, 3, 4 };
"nth" slist => { 3, 100, "-1", "-100" };

"test[$(nth)]" string => nth("test", $(nth));
```

Here only two strings will be generated because only two of the indices are valid:
```
test[3] = 4
test[-1] = 4
```

No errors, because the function just ignores indices that are out of bound.

nickanderson:
I will jus ttake this and followup with core example to illustrate

PR #2256, opened 2019-12-08
Re-parent chmmon promise attributes

nickanderson:
This moves the reference documentation for attributes that are available to all
promise types to it's own page called `Common` at the same level as individual
promise type documents, but sorted to be listed first.

----

#


craigcomstock:
Should we worry about old links to anchors like `https://docs.cfengine.com/docs/3.12/reference-promise-types.html#inherit_from` ?

nickanderson:
I would only make this change in master and 3.15

nickanderson:
But yes I need to search for links to these places first

nickanderson:
I don't think this will happen for 3.15.0, recinding

PR #2274, opened 2019-12-19
Updated supported platforms for 3.15.0

nickanderson:
- Dropped Enterprise Hub support on Ubuntu 14 and Debian 8
- Dropped standard support for EL5 and Debian 7
- Added standard support for EL8 (*except enterprise hubs)


nickanderson:
@roschern Can you please confirm?

roschern:
We should add that we drop standard support for SLES11.  We should support SLES12 and SLES15.

PR #2279, opened 2020-01-11
Specify different unpack for 32/64 bit tarball

lluiscampos:
The previous filename does not exist, the filename contains either i386 or x86_64.


cf-bottom:

PR #2280, opened 2020-01-16
ENT-5269: Added documentation about host decommissioning

olehermanse:


olehermanse:
@nickanderson I don't know if this is the best place to put this. Feel free to move it, or tell me where you think it belongs.

PR #2291, opened 2020-02-19
ENT-5266 Document base64-encoded reports

Lex-2008:


Lex-2008:
@nickanderson please check language/style

PR #2293, opened 2020-02-21
Add a section about FR HW requirements

vpodzime:
Not based on any exact research and measurements, but rather on
numbers we have seen in our FR tests and some small amounts of
data we got from our customers. See the ticket for more details.

Ticket: ENT-5132

### DIFF
--- a/enterprise-cfengine-guide/federated-reporting.markdown
+++ b/enterprise-cfengine-guide/federated-reporting.markdown
@@ -34,6 +34,8 @@ are provided at each stage of installation and setup that follows.
 
 ## Requirements ##
 
+### Software Requirements ###
+
 If your hub will have SELinux enabled, the `semanage` command must be installed.
 This allows Federated Reporting policy to manage the trust between the superhub and
 feeder hubs.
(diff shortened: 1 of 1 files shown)


enterprise-cfengine-guide/federated-reporting.markdown
```diff
@@ -51,6 +53,60 @@ federation policy to ensure that `semanage` is installed.
...
+would degrade the freshness of the data available on the Superhub.
+
+The recommended HW configuration for a Superhub with the default configuration
+and 5000 hosts per connected Feeder is:
+
+  * 16 GiB of RAM or more,
+
+  * 1 logical CPU per connected Feeder or more,
+
+  * 5 MiB of disk space per host or more,
```
nickanderson: I guess this is down from the 100MB per host we recommend on each hub because we don't aggregate monitoring data?

nickanderson:
Merged, backported to 3.15.

enterprise-cfengine-guide/federated-reporting.markdown
vpodzime: Yes. But also that 100 MiB per host is crazy, I think.

PR #2295, opened 2020-02-24
Fixed tiny grammar issue

craigcomstock:
"If it is close to the configured refresh interval" instead of "If it close to the configured refresh interval"


vpodzime:
Backported to 3.15

PR #2301, opened 2020-03-17
Document the new $(sys.cf_version_release) variable

vpodzime:
Added in cfengine/core@7005511250.


vpodzime:
Pushed to *3.15* as well.

PR #2307, opened 2020-04-03
Improved manual feeder setup slightly

craigcomstock:
re-ordered `sudo` arg to correct place for latest `cf-remote`.
fixed typo of `CLOUDUSER` versus `CLOUD_USER`

Changelog: none
Ticket: none


craigcomstock:
This should definitely go to 3.12, but what about 3.14 as well?

PR #2314, opened 2020-05-08
ENT-5965/master: Clarify how to control what is or isn't collected by cf-hub

nickanderson:
Ticket: ENT-5965
Changelog: None

----

#


nickanderson:
> LGTM, please remember to check if after it goes live.

Yep, looks good to me https://docs.cfengine.com/docs/master/enterprise-cfengine-guide-reporting.html

Backporting to 3.15 and 3.12

PR #2323, opened 2020-06-11
Unify commands and console snippets in examples

lluiscampos:
The criteria used is:
- When instructing the user to run a command, format it as console but
  do not prepend the console prompt.
- When showing the result of a command, format it also as console and
  prepend the prompt.

Minor adjustment in other snippet missing cf3 format.

Signed-off-by: Lluis Campos <lluis.campos@northern.tech>


lluiscampos:
The PR address current inconsistencies in examples page, see https://docs.cfengine.com/docs/3.15/examples.html

PR #2326, opened 2020-06-12
Apply Policy Style Guide to policy snippets in Promise Types reference

lluiscampos:


lluiscampos:
After fixing this markdown file I realized that the style is basically not followed anywhere. Before jumping to fix all the other files, I wanted to get feedback on how valuable this actually is...

PR #2327, opened 2020-06-15
Document files promise content attribute

lluiscampos:


reference/promise-types/files.markdown
```diff
@@ -1404,6 +1404,30 @@ comparison or linking operations).
...
+
+**Description:** Content to write into file
+
+**Type:** `string`
+
+**Allowed input range:** (arbitrary string)
+
+**Example:**
+
+```cf3
```
nickanderson, line 1417: Instead of embedding a snippet example like this, I thinkit's better to dynamically include the example from core.

https://github.com/cfengine/documentation/blob/master/README.md#quoting-policy-files

e.g. https://github.com/cfengine/documentation/blob/master/reference/promise-types/files.markdown#missing_ok

reference/promise-types/files.markdown
```diff
@@ -1404,6 +1404,30 @@ comparison or linking operations).
...
+
+  "/path/to/file"
+
+    content => "Hello world!";
+
+```
+
+**History:** Was introduced in 3.16.0
+
+**Note:** You cannot `content` in combination with the other edit operations
```
nickanderson, line 1428: What does the agent do if you do combine those attributes that are mutually exclusive?

lluiscampos: @nickanderson That is a good question. I am writing some acceptance tests for this. My idea was that `cf-agent` should log the error message and exit with error. Currently it will print something like:
```
error: A file promise with content attribute cannot have edit operations
```

reference/promise-types/files.markdown
lluiscampos: Thanks for this tip! I wasn't aware of it. I will rework.

reference/promise-types/files.markdown
lluiscampos: See acceptance tests at https://github.com/cfengine/core/pull/4219

reference/promise-types/files.markdown
lluiscampos: Fixed :+1:

reference/promise-types/files.markdown
```diff
@@ -1404,6 +1404,25 @@ comparison or linking operations).
 
 **See also:** [Common Body Attributes][Promise Types and Attributes#Common Body Attributes]
 
+### content
+
+**Description:** Content to write into file
```
nickanderson, line 1409: I think this should rather say `Complete content the file should contain`

lluiscampos:
> I just changed it directly myself.

@nickanderson Thanks! I squashed it now for nicer Git history. I am still waiting for the review in the implementation PR, anyway.

PR #2333, opened 2020-07-20
Added new regex operators to Inventory API documentation

aleksandrychev:
Ticket: ENT-6057
Merge together: https://github.com/cfengine/nova/pull/1690

### DIFF
--- a/reference/enterprise-api-ref/inventory.markdown
+++ b/reference/enterprise-api-ref/inventory.markdown
@@ -51,6 +51,9 @@ Inventory API allows to access inventory reports and attributes dictionary.
     | not_match   |
     | contains    |
     | not_contain |
+    | regex_matches |
+    | regex_not_match |
+    
 
 * **sort** *(string)*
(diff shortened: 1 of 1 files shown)


reference/enterprise-api-ref/inventory.markdown
```diff
@@ -51,6 +51,9 @@ Inventory API allows to access inventory reports and attributes dictionary.
     | not_match   |
     | contains    |
     | not_contain |
+    | [regex_matches](https://en.wikibooks.org/wiki/Regular_Expressions/POSIX-Extended_Regular_Expressions) |
```
craigcomstock, line 54: I'm not sure I would include a link. I think regex is easy enough to look up or just "know". :)

aleksandrychev: Ok, I will remove it.

aleksandrychev: @craigcomstock done

PR #2335, opened 2020-07-29
ENT-6100/master: Aligned supported platforms with download tables as of 3.15.2

nickanderson:


Lex-2008:
maybe let's merge it to 3.15 branch, instead, and keep master branch for 3.16 and following verisons?

nickanderson:
Well, we haven't made any platform changes yet. I thinik it's better to keep it accurate, then we can update master when we do decide.

nickanderson:
I won't backport to 3.16 docs because we dont have the branches and stuff to really support tweaking the docs

PR #2350, opened 2020-08-17
CFE-3402: Reverted parts of commit which trimmed newlines from images

olehermanse:
Bash oneliner:

```
git show 2a42e3158ddb7a2fc5098fe408d708c0ac19fab1 | grep "diff --git" | awk '{ print $4 }' | grep -E '.*\.(png|jpg)' | awk '{ print substr($1, 3) }' | xargs ls -d 2>/dev/null | gxargs -I @ sh -c 'printf "%s @ \n" "$(git log -n 1 --pretty=format:%H -- @)" ' | grep 2a42e3158ddb7a2fc5098fe408d708c0ac19fab1 | awk '{ print $2 }' | gxargs -I @ sh -c 'git checkout 2a42e3158ddb7a2fc5098fe408d708c0ac19fab1~1 @'
```

Of the pngs / jpegs in that commit, filter by those which still
exist and the most recent commit is that same commit. Checkout
those files at 1 commit before that SHA.

(The command above is by no means bullet proof!)

Ref:
cfengine/documentation@2a42e3158ddb7a2fc5098fe408d708c0ac19fab1


Lex-2008:
there is also at least one `*.gif` file

olehermanse:
> there is also at least one *.gif file

@aleksei https://github.com/cfengine/documentation/pull/2351

PR #2353, opened 2020-08-19
syntax error in body file control inputs example

craigcomstock:


cf-bottom:

reference/masterfiles-policy-framework/lib-common.markdown
```diff
@@ -15,7 +15,7 @@ To use these bodies, add the following to your policy:
 ```cf3
 body file control
 {
-	inputs => { "common.cf" }
+	inputs => { "common.cf" };
```
nickanderson, line 18: Instead of fixing this syntax error, should we instead redact the block and the instructions? common.cf is included by stdlib.cf by default.

craigcomstock: oh, right! Maybe I will look for other examples and see if they are "ok".

craigcomstock:
will open a different pull request with this "fixed" in a more thorough manner.

PR #2354, opened 2020-08-20
Add links to validjson/validdata to readjson/readdata respectively

vpodzime:
New functions that are related to the old ones.


vpodzime:
@nickanderson is this all that needs to be done to make these new *See also* links work?

nickanderson:
Yep

PR #2365, opened 2020-09-16
Positive and negative values do not match

amousset:
seems the correct value is eleven nines


cf-bottom:

nickanderson:
Indeed!

Good catch, thanks for fixing it.

nickanderson:
Picked back to 3.15 and 3.12

PR #2366, opened 2020-09-21
Extended Host API documentation

aleksandrychev:
ChangeLog: Extended Host API documentation with an example of getting hosts list grouped by hard classes
Ticket: ENT-6175

### DIFF
--- a/reference/enterprise-api-ref/host.markdown
+++ b/reference/enterprise-api-ref/host.markdown
@@ -126,6 +126,58 @@ future. Key purging usually take an effect within 5-10 minutes.
 of [`trustkeysfrom`][cf-serverd#trustkeysfrom] for the hub hosts may re-appear
 and resume being collected from after being deleted.
 
+## Hosts list grouped by hard classes
+
+**URI:** https://hub.cfengine.com/api/hosts/by-class
+
+**Method:** GET
+
+**Parameters:**
+
+* **context-include** *(comma delimited string of regular expression strings)*
+* **format** *(string)*
+    Output format. Default value is `json`. Allowed values: `json`, `yaml`.
+    
+**CURL unfiltered request example**
+```
+curl -k --user admin:admin -X GET https://hub.example.com/api/hosts/by-class
+```
+
+**Example response:**    
+```
+{
+    "10_0_2_15": [
+        "ubuntu-xenial.com"
+    ],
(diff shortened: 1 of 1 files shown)


reference/enterprise-api-ref/host.markdown
```diff
@@ -126,6 +126,39 @@ future. Key purging usually take an effect within 5-10 minutes.
 of [`trustkeysfrom`][cf-serverd#trustkeysfrom] for the hub hosts may re-appear
 and resume being collected from after being deleted.
 
+## Hosts list grouped by hard classes
+
+**URI:** https://hub.cfengine.com/api/hosts/by-class
```
nickanderson, line 131: Still feels to me like our api endpoints should be versioned.

reference/enterprise-api-ref/host.markdown
```diff
@@ -126,6 +126,39 @@ future. Key purging usually take an effect within 5-10 minutes.
...
+**Method:** GET
+
+**Parameters:**
+
+* **context-include** *(comma delimited string of regular expression strings)*
+    Includes hosts having context matching the expression.
+* **format** *(string)*
+    Output format. Default value is `json`. Allowed values: `json`, `yaml`.
+
+**Example response:**
```
nickanderson, line 142: Can you please add in a CURL example call?

`curl -u USER:TOKEN https://hub/api/hosts/by-class?context-include=10_.*,127_.*,172_,ubuntu_.*`
?

nickanderson (changes requested):
Looks pretty good, can you please include an example call for curl?

reference/enterprise-api-ref/host.markdown
aleksandrychev: From my point of view, versioning makes no too much sense right now because all changes are backward compatible. we might add versioning when some breakable changes gonna be added. And having version as part of the URL isn't the best practice of versioning.

reference/enterprise-api-ref/host.markdown
aleksandrychev: yes, sure.

aleksandrychev: done

craigcomstock (approved):

nickanderson (approved):

PR #2367, opened 2020-09-24
fix endraw tag

kkaempf:
The `endraw` tag is wrong and leads to the following error when running `jekyll build`:

  Liquid Exception: Liquid syntax error (line 518): 'raw' tag was never closed in cheatsheet.markdown


cf-bottom:

nickanderson:
Thank you!

PR #2369, opened 2020-09-24
CFE-3427/master: Changed references from bundle server access_rules to my_access_rules

nickanderson:
This change is intended to help clarify the fact that users can (and should)
define their own server bundles instead of modifying the vendored policy and
that bundle server access_rules is not a special name.

Merge with: https://github.com/cfengine/masterfiles/pull/1839

### DIFF
--- a/reference/components/cf-serverd.markdown
+++ b/reference/components/cf-serverd.markdown
@@ -341,48 +341,50 @@ The sequence of events is this:
 
     call_collect_interval => "5";
 
-The full configuration would look something like this
+The full configuration to enable client initiated reporting would look something like this:
 
 ```cf3
-        #########################################################
-        # Server config
-        #########################################################
+#########################################################
+# Server config
+#########################################################
 
-        body server control
-        {
-        allowconnects         => { "10.10.10.0/24" , "::1" };
-        allowallconnects      => { "10.10.10.0/24" , "::1" };
-        trustkeysfrom         => { "10.10.10.0/24" , "::1" };
+body server control
+{
+  allowconnects         => { "10.10.10.0/24" , "::1" };
+  allowallconnects      => { "10.10.10.0/24" , "::1" };
+  trustkeysfrom         => { "10.10.10.0/24" , "::1" };
 
-        call_collect_interval => "5";
-        }
+  call_collect_interval => "5";
+}
 
-        #########################################################
+#########################################################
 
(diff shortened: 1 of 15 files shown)


nickanderson:
Ping @basvandervlies to clarify that bundle server access_rules isn't anything unique and special.

reference/components/cf-serverd.markdown
```diff
@@ -344,44 +344,43 @@ The sequence of events is this:
 The full configuration would look something like this
 
 ```cf3
-        #########################################################
-        # Server config
-        #########################################################
+#########################################################
```
craigcomstock, line 347: This is the only section that made me pause. Which "full configuration" file would this be in? If you were modifying controls/cf_execd.cf then you probably wouldn't change to my_access_rules bundle name. If you were creating a replacement you would need to change promises.cf inputs[cf_execd].

basvandervlies:
Thanks @nickanderson for me it is now clear  how to use it. I will try it in out configuration

reference/components/cf-serverd.markdown
nickanderson: OK, I have tweaked it very slightly. This part of the documentation is about the configuration for collect_calls access. This "full configuration" snippet shows all of the various settings related to enabling client-initiated reporting. It's not specific to MPF. If you have any suggestions to clarify further please let me know. It's challenging to straddle the difference between documentation about core, and then conventions from MPF which won't necessarily apply if you run a completely custom policy set.

craigcomstock: Cool. I think for me I would like to see the text before the example say: change controls/cf_serverd.cf as follows (and then the two examples), or if my_access_rules() is really just augmenting what is in cf_server.cf, which it is, then you might break that section apart and prefix it with some text like: include this `bundle server my_access_rules()` in a separate file and include in inputs with augments (def.json). 

Something like that make any sense to you?

nickanderson: So, the access rules are already present by default, and we have augments about them.

Again, this documetation about the control body is about the core component with has nothing really to do with MPF. I could however add some See Also references to the MPF augments for these settings.

olehermanse (approved):

olehermanse (changes requested):
LGTM, please squash commits.

PR #2378, opened 2020-10-13
CFE-3462/master: Added documentation about special cases in which variables are defined in foreign bundles

nickanderson:
Ticket: CFE-3462
Changelog: None

### DIFF
--- a/reference/promise-types/vars.markdown
+++ b/reference/promise-types/vars.markdown
@@ -378,3 +378,272 @@ two_example_com::
 ```
 
 (Promises within the same bundle are evaluated top to bottom, so vars promises further down in a bundle can overwrite previous values of a variable. See [**normal ordering**][Normal Ordering] for more information).
+
+## Defining variables in foreign bundles
+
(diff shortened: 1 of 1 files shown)


reference/promise-types/vars.markdown
```diff
@@ -378,3 +378,144 @@ two_example_com::
 ```
 
 (Promises within the same bundle are evaluated top to bottom, so vars promises further down in a bundle can overwrite previous values of a variable. See [**normal ordering**][Normal Ordering] for more information).
+
+## Defining variables in foreign bundles
+
+As a general rule, variables can only be defined from within the bundle where a
+variable is defined. There are a few notable exceptions to this general rule
```
vpodzime, line 385: The first sentence doesn't quite make sense.

vpodzime (commented):
Looks good to me otherwise.

reference/promise-types/vars.markdown
nickanderson: Updated

PR #2400, opened 2021-01-18
Fix inline YAML example

tzz:


reference/promise-types/vars.markdown
```diff
@@ -215,12 +215,12 @@ For example:
...
       # JSON or YAML can be inlined since CFEngine 3.7
       "inline1" data => '{"key":"value"}'; # JSON
-      "inline2" data => '---\n- key2: value2'; # YAML requires "---\n" header
+      "inline2" data => '---$(const.n)- key2: value2'; # YAML requires "---\n" header
 ```
 
 Inline YAML data has to begin with the `---\n` preamble. This preamble
 is normally optional but here (for inline YAML) it's required by
-CFEngine.
+CFEngine. To generate that in CFEngine, use `$(const.n)`.
```
atsaloli, line 223: Thank you for proposing this change, Ted.

Should we also mention that one can also inline the newlines directly?  That might make the resulting code more readable.

For example:

image

Although I suppose for really long YAML documents one would be better off using `readyaml()` instead.

I just don't want to give the impression that one HAS to use `$(const.n)`.

That could be the subject of a separate PR, it doesn't have to be fixed here (to keep us from scope drift).  :-)  I certainly don't object to this change.

tzz: Good call, edited. Thank you!

nickanderson (approved):
FYI, I opened CFE-3551 related to this.

PR #2415, opened 2021-03-04
Update HA installation instructions to RHEL/CentOS 7

vpodzime:
RHEL/CentOS 6 is EOL so we shouldn't encourage users to build
their HA setups on it.

Ticket: ENT-6825
Changelog: None

### DIFF
--- a/enterprise-cfengine-guide/high-availability/installation-guide.markdown
+++ b/enterprise-cfengine-guide/high-availability/installation-guide.markdown
@@ -24,7 +24,7 @@ all your CFEngine clients in case of failover.
 
 ### Hardware configuration and OS pre-configuration steps ###
 
-* CFEngine 3.6.2 (or later) hub package for RHEL6 or CentOS6.
+* CFEngine 3.15.3 (or later) hub package for RHEL7 or CentOS7.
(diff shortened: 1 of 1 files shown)


enterprise-cfengine-guide/high-availability/installation-guide.markdown
```diff
@@ -503,7 +513,7 @@ HA fencing guide](https://access.redhat.com/documentation/en-us/red_hat_enterpri
 
    **IMPORTANT:** Copy over only the hashes, without the `SHA=` prefix.
 
-5. **On both nodes,** modify the */var/cfengine/masterfiles/controls/def.cf* and
+6. **On both nodes,** modify the */var/cfengine/masterfiles/controls/def.cf* and
```
nickanderson, line 516: Instead of instructing people to modify two policy files, I think we should instead suggest defining the class via augments.

```
jq '.classes +=  { "enable_cfengine_enterprise_hub_ha": [ "any::" ] }' < masterfiles/def.json > masterfiles/def.json.new
mv masterfiles/def.json{.new,}
```

nickanderson (changes requested):

nickanderson (approved):

PR #2444, opened 2021-04-23
Added CMDB API documentation

aleksandrychev:
Ticket: ENT-6785

ChangeLog: Title

### DIFF
--- a/reference/enterprise-api-ref/cmdb-api.markdown
+++ b/reference/enterprise-api-ref/cmdb-api.markdown
@@ -19,17 +19,17 @@ You can see a list of stored host-specific configurations
 * **fromEpoch** *(integer)*
   Returns configurations with epoch value greater than set in the filter.
   Epoch is the sequence number of the latest CMDB change. In every API list request,
-  `most_recent_cmdb_change` will be present in the meta section, which contains the maximum
+  `cmdb_epoch` will be present in the meta section, which contains the maximum
   epoch value among selected items. Optional parameter.
 * **fromTime** *(timestamp)*
-  Include changes performed within interval. Format: `YYYY-mm-dd HH:MM:SS` or `YYYY-mm-dd`
+  Include changes performed within interval. Format: `YYYY-mm-dd HH:MM:SS` or `YYYY-mm-dd`. Optional parameter.
 * **toTime** *(timestamp)*
-  Include changes performed within interval. Format: `YYYY-mm-dd HH:MM:SS` or `YYYY-mm-dd`
+  Include changes performed within interval. Format: `YYYY-mm-dd HH:MM:SS` or `YYYY-mm-dd`. Optional parameter.
 * **skip** *(integer)*
   Number of results to skip for the processed
   query. The Mission Portal uses this for pagination. Optional parameter.
 * **limit**  *(integer)*
-  Limit the number of results in the query.
+  Limit the number of results in the query. Optional parameter.
 * **hostContextInclude** *(array)*
   Includes only results that concern hosts which have all specified CFEngine contexts (class) set. Optional parameter.
 * **hostContextExclude** *(array)*
@@ -41,8 +41,7 @@ You can see a list of stored host-specific configurations
 ```
 curl -k --user <username>:<password> \
   -X GET \
-  https://hub.cfengine.com/api/cmdb?epochFrom=2&hostContextInclude[]=ubuntu \
-  -H 'content-type: application/json' 
+  https://hub.cfengine.com/api/cmdb?epochFrom=2&hostContextInclude[]=ubuntu 
 ```
 
 **Example response:**
@@ -74,12 +73,54 @@ HTTP 200 Ok
         "page": 1,
         "timestamp": 1619116399,
         "total": "1",
-        "most_recent_cmdb_change": "13"
+        "cmdb_epoch": "13"
     }
 }
 ```
 
-## Get configuration by hostkey
+## Get host's specific configuration
+
+**URI:** https://hub.cfengine.com/cmdb/:hostkey/:type/:name/
+
+**Method:** GET
+
+**Parameters:**
+* **hostkey** *(string)*
+  Unique host identifier.
+
+* **type** *(string)*
+  Configuration type. Allowed value: `variables`, `classes`
+
+* **name** *(string)*
+  Configuration name. Classes or variables name.  
+
+**Example request (curl):**
+
+```
+curl -k --user <username>:<password> \
+  -X GET \
+  https://hub.cfengine.com/api/cmdb/SHA=f622992fa4525070f47da086041a38733496f03a77880f70b1ce6784c38f79ab/variables/HubCMDB:My.hostname/
+```
+
+**Example response:**
+
+```
+HTTP 200 Ok
+{
+    "hostkey": "SHA=437d63cdc0b13ad18bb2d9de2490bfabe4edc8aa59f248b5b5b050c77bf4eeef",
+    "variables": {
+        "default:def.augment_inputs": {
+            "tags": [
+                "suggestion-004"
+            ],
+            "value": [],
+            "comment": "Add filenames to this list to make the CFEngine agent parse them. Note: Update the bundle sequence to evaluate bundles from these policy files."
+        }
+    }
+}
+```
+
+## Get host's configurations
 
 **URI:** https://hub.cfengine.com/cmdb/:hostkey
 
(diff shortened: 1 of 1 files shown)


reference/enterprise-api-ref/cmdb-api.markdown
```diff
@@ -0,0 +1,317 @@
...
+title: CMDB API
+published: true
+tags: [reference, enterprise, API, CMDB, classes, variables]
+---
+
+Configuration management database API enables you to manage classes and variables for specific hosts.
+
+## List CMDB 
+
+You can see a list of stored hosts-specific configurations
```
craigcomstock, line 12: maybe s/hosts/host/ ? singular?

reference/enterprise-api-ref/cmdb-api.markdown
```diff
@@ -0,0 +1,317 @@
...
+
+**URI:** https://hub.cfengine.com/cmdb
+
+**Method:** GET
+
+**Parameters:**
+* **fromEpoch** *(integer)*
+  Returns configurations with epoch value greater than set in the filter.
+  Epoch is the sequence number of the latest CMDB change. In every API list request, you
+  might see `most_recent_cmdb_change` property on the meta section, which contains the maximum
```
craigcomstock, line 22: why "might"? Shouldn't it either be "will" or not? 

s/on the meta/in the meta/

reference/enterprise-api-ref/cmdb-api.markdown
```diff
@@ -0,0 +1,317 @@
...
+                }
+            }
+        }
+    },
+    "meta": {
+        "count": "1",
+        "page": 1,
+        "timestamp": 1619116399,
+        "total": "1",
+        "most_recent_cmdb_change": "13"
```
craigcomstock, line 77: looking at this here I don't like that "most_recent_cmdb_change" has no obvious link to "epoch". That change would be in code as well but it would make more sense to me to call this "cmdb_epoch".

reference/enterprise-api-ref/cmdb-api.markdown
```diff
@@ -0,0 +1,317 @@
...
+  Example:
+```
+{
+   "classes":{
+      "My_class":{
+         
+      },
+      "My_class2":{
+         "comment":"comment body"
+      }
```
craigcomstock, line 147: maybe include also a variables section here?

reference/enterprise-api-ref/cmdb-api.markdown
```diff
@@ -0,0 +1,317 @@
...
+  
+* **variables** *(JSON object)*
+  The format is a JSON object where the key is variable name and value is another JSON object
+  with a required `value` property and an optional `comment` property.
+  Example:
+```
+{
+   "variables":{
+      "Namespace:BundleName.VariableName":{
+         "value":"myvalue"
```
craigcomstock, line 160: Why are there two separate examples here? Do you really have to specify a json object that has the key "variables" in the value given for the variables parameter? This seems redundant and kind of odd.

reference/enterprise-api-ref/cmdb-api.markdown
```diff
@@ -0,0 +1,317 @@
...
+**Example request (curl):**
+
+```
+curl -k --user <username>:<password> \
+  -X POST \
+  https://hub.cfengine.com/api/cmdb \
+  -H 'content-type: application/json' \
+  -d '{
+   "hostkey":"SHA=f622992fa4525070f47da086041a38733496f03a77880f70b1ce6784c38f79ab",
+   "classes":{
```
craigcomstock, line 180: Ah, I see. The previous examples were snippets. Hmm. Do you think you could re-do this to make it more obvious?

reference/enterprise-api-ref/cmdb-api.markdown
```diff
@@ -0,0 +1,317 @@
...
+```
+
+## Update configuration
+
+**URI:** https://hub.cfengine.com/cmdb/:hostkey
+
+**Method:** PATCH
+
+**Parameters:**
+* **hostkey** *(string)*
```
craigcomstock, line 211: same as above, it seems a bit confusing to me to include the classes and variables "keys" in the examples when the value for those parameters is really just the internal json structure instead. So maybe remove the key part in the snippets like:

classes example:
```
{
  "My_class": {},
  "My_class2":{
    "comment":"comment body"
  }
}
```

reference/enterprise-api-ref/cmdb-api.markdown
nickanderson: In the data that the agent loads, yes _variables_ is the key that the definitions must be under.

For modifying data in the cmdb we don't have to re-use this format exactly, but I guess that it's required here since you could set vars and classes in the cmdb concurrently

reference/enterprise-api-ref/cmdb-api.markdown
nickanderson: In order to remove the variables and classes keys from the POST I think we would have to then have 2 different endpoints, one for vars and one for classes.

reference/enterprise-api-ref/cmdb-api.markdown
```diff
@@ -0,0 +1,317 @@
...
+   }
+}
+```
+If you need to delete all variables from host you need to set null value:
+```
+{
+    "variables": null
+}
+```
+If your request body misses variables then the previous value will be preserved.
```
nickanderson, line 261: This makes me wonder, do I have to provide _all_ previously defined variables when setting a new one?

reference/enterprise-api-ref/cmdb-api.markdown
craigcomstock: I wasn't saying remove it from the data sent to the end-point, but rather remove it from the snippets and then put it together later. It was confusing to me.

craigcomstock: re-reading it now it makes sense, but initially it was confusing. Maybe ask another non-experienced person to review this?! @aleksandrychev

reference/enterprise-api-ref/cmdb-api.markdown
nickanderson: @olehermanse @vpodzime care to comment on the colour of this bike shed?

reference/enterprise-api-ref/cmdb-api.markdown
nickanderson: The example request below includes both.

reference/enterprise-api-ref/cmdb-api.markdown
nickanderson: @craigcomstock you still think this is redundant? to have a json illustrating classes and vars separately, with the example showing both?

reference/enterprise-api-ref/cmdb-api.markdown
nickanderson: ping @olehermanse  @chiachenglu

reference/enterprise-api-ref/cmdb-api.markdown
nickanderson: @aleksandrychev ?

aleksandrychev: in this case, yes, but since I opened this PR I've added new API calls to add/update/delete specific items. I will add them later this week.

reference/enterprise-api-ref/cmdb-api.markdown
```diff
@@ -0,0 +1,317 @@
...
+**Parameters:**
+* **fromEpoch** *(integer)*
+  Returns configurations with epoch value greater than set in the filter.
+  Epoch is the sequence number of the latest CMDB change. In every API list request, you
+  might see `most_recent_cmdb_change` property on the meta section, which contains the maximum
+  epoch value among selected items. Optional parameter.
+* **fromTime** *(timestamp)*
+  Include changes performed within interval. Format: `YYYY-mm-dd HH:MM:SS` or `YYYY-mm-dd`
+* **toTime** *(timestamp)*
+  Include changes performed within interval. Format: `YYYY-mm-dd HH:MM:SS` or `YYYY-mm-dd`
```
vpodzime, line 27: Aren't these two also optional?

reference/enterprise-api-ref/cmdb-api.markdown
```diff
@@ -0,0 +1,317 @@
...
+* **skip** *(integer)*
+  Number of results to skip for the processed
+  query. The Mission Portal uses this for pagination. Optional parameter.
+* **limit**  *(integer)*
+  Limit the number of results in the query.
```
vpodzime, line 32: And this one too?

reference/enterprise-api-ref/cmdb-api.markdown
```diff
@@ -0,0 +1,317 @@
...
+  Excludes results that concern hosts which have specified CFEngine context (class) set. Hosts that have at least one of the specified contexts set will be excluded from the results. Optional parameter.
+
+    
+**Example request (curl):**
+ 
+```
+curl -k --user <username>:<password> \
+  -X GET \
+  https://hub.cfengine.com/api/cmdb?epochFrom=2&hostContextInclude[]=ubuntu \
+  -H 'content-type: application/json'
```
vpodzime, line 45: Is the `content-type` header necessary?

reference/enterprise-api-ref/cmdb-api.markdown
vpodzime: `cmdb_epoch` makes more sense to me given the `fromEpoch` argument.

reference/enterprise-api-ref/cmdb-api.markdown
```diff
@@ -0,0 +1,317 @@
...
+**URI:** https://hub.cfengine.com/cmdb
+
+**Method:** POST
+
+**Parameters:**
+* **hostkey** *(string)*
+  Unique host identifier.
+* **classes** *(JSON object)*
+  The format is a JSON object where the key is class name and value is another JSON object
+  with an optional `comment` property.
```
vpodzime, line 137: So `tags` are not supported by the API? (and the MP side of CMDB in general?)

vpodzime (commented):
Looks good to me otherwise.

reference/enterprise-api-ref/cmdb-api.markdown
aleksandrychev: renamed to `cmdb_epoch`

reference/enterprise-api-ref/cmdb-api.markdown
aleksandrychev: supported, added here

reference/enterprise-api-ref/cmdb-api.markdown
aleksandrychev: yeah, not needed, removed

reference/enterprise-api-ref/cmdb-api.markdown
aleksandrychev: nope, they aren't. added that

PR #2448, opened 2021-05-10
CFE-3616: Added int() policy function documentation

KotzaBoss:
Ticket: CFE-3616
Changelog: None

### DIFF
--- a/reference/functions/int.markdown
+++ b/reference/functions/int.markdown
@@ -1,0 +1,24 @@
+---
+layout: default
+title: int
+published: true
+tags: [reference, functions, int]
+---
+
+[%CFEngine_function_prototype(string)%]
+
+**Description:** Convert numeric string to int.
+
+[%CFEngine_function_attributes(string)%]
+
+If `string` represents a floating point number then the decimals are *truncated*.
+
+**Example:**
+
+[%CFEngine_include_example(int.cf)%]
+
+**See Also:** [`string()`][string]
+
+**History:**
+
+* Introduced in 3.18.0


reference/functions/int.markdown
```diff
@@ -0,0 +1,44 @@
...
+
+[%CFEngine_function_prototype(string)%]
+
+**Description:** Convert numeric string to int.
+
+[%CFEngine_function_attributes(string)%]
+
+If `string` represents a floating point number then the decimals are *truncated*.
+
+```cfengine3
```
nickanderson, line 16: Instead of having an inline example, I think we should leverage your the example from core.

Replace all that with:

  `[%CFEngine_include_example(int.cf)%]`

reference/functions/int.markdown
```diff
@@ -0,0 +1,44 @@
...
+Output:
+
+```
+R: "123.456" -> 123
+R: "567.789" -> 567
+```
+
+**Example:**
+
+[%CFEngine_include_example(int.cf)%]
```
nickanderson, line 41: Oh, you have an example here, so I think you should delete the embedded example and output.

reference/functions/int.markdown
```diff
@@ -0,0 +1,44 @@
...
+
+**See Also:** [`string()`][string]
+
```
nickanderson, line 44: And we should also have a History section.
```
**History:**

* Introduced in 3.18.0
```

nickanderson (changes requested):
Looks nice, just a couple of changes requested.

PR #2450, opened 2021-05-11
CFE-2240: Added documentation on policy function `type()`

larsewi:
Ticket: CFE-2240
Changelog: None
Signed-off-by: larsewi <lars.erik.wik@northern.tech>


nickanderson (commented):
I will make some modifications after merge, so just be sure to come back and review the end state.

PR #2451, opened 2021-05-11
CFE-3558: Update reports promise doc with recent changes to printfile body

larsewi:
Updated doc to illustrate how to print from end of file with
`number_of_lines` set to a negative number in printfile body.

Ticket: CFE-3558
Changelog: None
Signed-off-by: larsewi <lars.erik.wik@northern.tech>

### DIFF
--- a/reference/promise-types/reports.markdown
+++ b/reference/promise-types/reports.markdown
@@ -112,24 +112,20 @@ Include part of a file in a report.
 
 **Default value:** `5`
 
-**Allowed input range:** `0,99999999999`
+**Allowed input range:** `-99999999999,99999999999`
+
+**Note:** Prints lines from end of file when a negative number is passed to
+argument `number_of_lines`
 
 **Example:**
 
-```cf3
-     bundle agent example
-     {
-     reports:
-         "$(sys.date) - current message of the day:"
-            printfile => "motd";
-     }
+[%CFEngine_include_snippet(printfile.cf, #\+begin_src prep, .*end_src)%]
 
-     body printfile motd
-     {
-         file_to_print   => "/etc/motd";
-         number_of_lines => "10";
-     }
-```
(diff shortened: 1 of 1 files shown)


reference/promise-types/reports.markdown
```diff
@@ -112,24 +112,20 @@ Include part of a file in a report.
 
 **Default value:** `5`
 
-**Allowed input range:** `0,99999999999`
+**Allowed input range:** `-99999999999,99999999999`
```
nickanderson, line 115: hurmmmmmm.
Feels like this should be a macro.

nickanderson: We have function prototype macro, and we have a promise attribute macro, perhaps we lack a body attribute macro.

nickanderson: scanning the code it looks like promise_attribute knows about body attributes.

```
	if header[2].find("Attributes") != -1:
		# assumption for promise type definition
		# header[1] = promise type
		# header[2] = "Attributes"
		# header[3] = attribute name
		# header[4] = body attribute
		promise_type_def = promise_types[header[1]]
		attribute_def = promise_type_def["attributes"][header[3]]
		attribute_type = attribute_def["type"]
		if attribute_type == "body":
			if header[4]:
				body_type_def = body_types[header[3]]
				attribute_def = body_type_def["attributes"][header[4]]
				attribute_type = attribute_def["type"]
			else:
				lines.append("**Type:** `body %s`\n\n" % header[3])
				return lines

```

Sooooooo, I think we should try 

```
[%CFEngine_promise_attribute(5)%]
```
And that hopefully then renders Type, default value and allowed input range.

nickanderson (changes requested):

reference/promise-types/reports.markdown
nickanderson: It didnt :(

PR #2470, opened 2021-06-17
Update augments.markdown

craigcomstock:
fixup some history sections which weren't getting formatted properly

### DIFF
--- a/reference/language-concepts/augments.markdown
+++ b/reference/language-concepts/augments.markdown
@@ -409,7 +409,7 @@ R: def.my_other_var == Defined ONLY in def.json
 R: def.centos_6_var == Defined ONLY in centos_6.json
 ```
 
-**History:**
+## History
 
 * 3.18.0
   * Introduced `variables` key with support for metadata (`comment`, `tags`) and targeting the _namespace_ and _bundle_.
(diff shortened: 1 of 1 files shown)


reference/language-concepts/augments.markdown
```diff
@@ -409,7 +409,7 @@ R: def.my_other_var == Defined ONLY in def.json
 R: def.centos_6_var == Defined ONLY in centos_6.json
 ```
 
-# History
+**History:**
```
nickanderson, line 412: Nope, it needs to be a heading or it only applies to the last section which is augments. It could be a h2 instead of h1 to put it at the same level as each of the keys.

reference/language-concepts/augments.markdown
```diff
@@ -358,8 +358,8 @@ myclass_defined_by_augments_in_def_json_3_18_0_v1            optional,tags,sourc
 **History:**
 
 * 3.18.0
-** Support for dict structure for classes and support for metadata (`comment`, `tags`) added.
-** Classes are defined as _soft_ classes instead of _hard_ classes.
+  * Support for dict structure for classes and support for metadata (`comment`, `tags`) added.
+  * Classes are defined as _soft_ classes instead of _hard_ classes.
```
nickanderson, line 362: yeah, i did that in anotehr section but must have forgotten this one.

PR #2475, opened 2021-06-25
ENT-7355 fix upgrading link in guide

Lex-2008:
The page is called just "upgrading" now, without version.


Lex-2008:
cherry-picked to 3.18, 3.15, and 3.12 branches

PR #2554, opened 2021-10-04
Fix links

rjc:
- _http_ `->` _https_ where appropriate and there's already an autoredirect in place
- prepend `docs.` where redirect already does so by default
- add missing `docs/`
- dashes (`-`) instead of underscores (`_`), and lowercase letters are used in links on the same page

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,9 +1,9 @@
 # CFEngine Documentation
 
 This repository holds the sources for the technical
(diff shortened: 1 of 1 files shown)


olehermanse (approved):
Looks good to me, thanks :)

nickanderson (approved):
Hey, thanks @rjc. Much appreciated.

merging, backporting.

PR #2583, opened 2021-10-29
Added VCS type to VCS settings API documentantion

aleksandrychev:


nickanderson (approved):
I added a History section and default value for vcsType, otherwise, seems ok.

PR #2645, opened 2022-02-25
Fixed vagrant ip for better compatibility

nickanderson:
Bringing up the vagrant environment with the provided config results in messages
like this:

==> hub: You assigned a static IP ending in ".1" to this machine.
==> hub: This is very often used by the router and can cause the
==> hub: network to not work properly. If the network doesn't work
==> hub: properly, try changing this IP.

Indeed, while the vagrant env got the proper ssh keys inserted I was unable to
log in directly with ssh (but could use vagrant ssh from within the project).

I recall that I had similar issues when originally making the vagrant project
years ago. Indeed, switching to .2 resolved the issue.

### DIFF
--- a/guide/getting-started-with-cfengine-build/installation.markdown
+++ b/guide/getting-started-with-cfengine-build/installation.markdown
@@ -128,7 +128,7 @@ For example, in Digital Ocean, the username is `root`, and the IP might be `128.
 
 ![](digital-ocean.png)
 
-**Note:** In the rest of this tutorial, replace the IP address we use in the examples, `192.168.56.2` with that IP.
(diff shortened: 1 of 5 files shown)


olehermanse:
@nickanderson I already switched docs to `192.168.56.2` for this reason, I don't think your change to `192.0.2.2` will improve things.

olehermanse:
The reason for using `56` is here:

> On Linux, Mac OS X and Solaris Oracle VM VirtualBox will only allow IP addresses in 192.168.56.0/21

https://www.virtualbox.org/manual/ch06.html#network_hostonly

nickanderson:
Ah good to know. We should make a change to the other packages virtual box environment.

PR #2652, opened 2022-03-03
CFE-951: Changed default value for rxdirs

nickanderson:
Ticket: CFE-951
Changelog: None


vpodzime (approved):
ACK

vpodzime:
@nickanderson do you think we need something on the *3.18* branch as well?

PR #2701, opened 2022-04-26
ENT-8560: Updated docs on protocol_version (3.15)

nickanderson:
We neglected to update the documentation after releasing protocol_version 3,
this is just back-filling those details.


cf-bottom:

PR #2713, opened 2022-06-09
CFE-160: Improved documentation for Namespaces

nickanderson:


reference/language-concepts/namespaces.markdown
```diff
@@ -6,104 +6,76 @@ sorting: 100
...
-`default` namespace can be accessed like any other:
+Bodies are assumed to be within the same namespace as the promiser. To use a body from another namespace the namespace must be specified by prefixing the body name with the namespace followed by a colon (`:`).
 
-```cf3
-    files:
-      "/file"
-         create => "true",
-          perms => default:settings;
-```
+A common mistake is forgetting to specify `default:` when using bodies from the standard library which resides in the `default` namespace.
```
craigcomstock, line 45: This kind of mentions what I was saying earlier but I think you could more clearly link the inclusion of a namespace in `body file control` to everything after needing `default:` to use standard library items.

reference/language-concepts/namespaces.markdown
```diff
@@ -6,104 +6,76 @@ sorting: 100
...
+Variables (except for Special Variables) are assumed to be within the same scope
+as the promiser but can also be referenced fully qualified with the namespace.
 
-**Note** that this means that if you are in a namespace that's not `default`, you *must* qualify classes from `default` fully:
+[%CFEngine_include_example(namespace_variable_references.cf)%]
 
-    default:myclass::
-    "do something" if => "default:myotherclass";
+[Special variables][Special Variables] are always accessible without a namespace
+  prefix. For example, `this`, `mon`, `sys`, and `const` fall in this category.
```
craigcomstock, line 57: how odd. But yes, good to mention. :)

craigcomstock (commented):
cool, nice revisions, a few questions...

reference/language-concepts/namespaces.markdown
```diff
@@ -6,104 +6,76 @@ sorting: 100
 tags: [manuals, language, syntax, concepts, namespace]
 ---
 
-Namespaces are private bundle and body "playgrounds", allowing
-multiple files to define the bundles and bodies with the same name in
-different namespaces without conflict.  They are key to writing
-reusable policies.
+By default all things are in the `default` namespace. Specifying a namespace
```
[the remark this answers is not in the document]
nickanderson: ACK

reference/language-concepts/namespaces.markdown
```diff
@@ -6,104 +6,76 @@ sorting: 100
...
+## Declaration
 
-To isolate a file into its own namespace, you add a
-[file control promise][file control#namespace] to the file before the relevant
-bundles or bodies. All bundles and bodies start off in the `default` namespace
-if you don't explicitly set this. Once set, this applies until the end of the
-file or the next namespace change.
+Namespaces are declared with [`body file control`][file control#namespace]. A
+namespace applies within a single file to all subsequently defined bodies
+following the namespace declaration until a different namespace has been
```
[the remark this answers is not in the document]
nickanderson: ACK, and regarding default: yes, it's mentoned in following sections, especially bodies where it's most commonly forgotten for referencing stdlib.

Examples also have some comments about it.

reference/language-concepts/namespaces.markdown
```diff
@@ -6,104 +6,76 @@ sorting: 100
...
 
-```cf3
-    files:
-      "/file"
-         create => "true",
-         action => if_repaired("done");
-```
+- The [`variables` Augments key][Augments#variables] defines variables in the
+  `main` bundle of the `data` namespace by default supports seeding variable
+  values in any specified namespace.
```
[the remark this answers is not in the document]
nickanderson: ack, missing but.

reference/language-concepts/namespaces.markdown
nickanderson: Yeah, also embedded in some examples: https://github.com/cfengine/core/pull/4966

if you have a differnnt concrete suggestion happy to amend the text

reference/language-concepts/namespaces.markdown
```diff
@@ -6,104 +6,76 @@ sorting: 100
...
-specify this `default:` prefix.
+[%CFEngine_include_example(namespace_bodies.cf)%]
 
-To access classes, variables, or meta-data in bundles in a different namespace, use the
-colon as a namespace prefix:
+## Variables
 
-    $(namespace:bundle.variable)
-    $(namespace:bundle_meta.variable)
+Variables (except for Special Variables) are assumed to be within the same scope
```
craigcomstock, line 51: s/scope/namespace/ , the term scope is not used any where else in this doc so using "namespace" will be more clear.

reference/language-concepts/namespaces.markdown
```diff
@@ -6,104 +6,76 @@ sorting: 100
...
+as the promiser but can also be referenced fully qualified with the namespace.
 
-**Note** that this means that if you are in a namespace that's not `default`, you *must* qualify classes from `default` fully:
+[%CFEngine_include_example(namespace_variable_references.cf)%]
 
-    default:myclass::
-    "do something" if => "default:myotherclass";
+[Special variables][Special Variables] are always accessible without a namespace
+  prefix. For example, `this`, `mon`, `sys`, and `const` fall in this category.
```
craigcomstock, line 57: actually here it might be good to say something about how these examples are already a prefix but with a dot? How does that work exactly? Seems like they are more like bundles/bodies these special variables?

reference/language-concepts/namespaces.markdown
```diff
@@ -6,104 +6,76 @@ sorting: 100
...
 
-```cf3
-    files:
-      "/file"
-         create => "true",
-         action => if_repaired("done");
-```
+- The [`variables` Augments key][Augments#variables] defines variables in the
+  `main` bundle of the `data` namespace by default but supports seeding variable
+  values in any specified namespace.
```
craigcomstock, line 65: an example would be nice just to be clear: aka variables item foo would be referred to as data:main.foo right?

reference/language-concepts/namespaces.markdown
```diff
@@ -6,104 +6,76 @@ sorting: 100
...
-Similarly, variables you create in a namespaced bundle have to be
-prefixed like `mynamespace:mybundle.myvar` from outside your
-namespace, but can use `mybundle.myvar` inside the namespace and
-`myvar` inside `mybundle`.
+## Classes
 
-As a workaround, you could have a helper bundle in another namespace
-to create classes and variables as needed.
+Promises can only define classes within the current namespace. Classes are
+understood to refer to classes in the current namespace if a namespace is not
```
craigcomstock, line 70: by "current namespace" here do you mean "default" or a special magical namespace that exists for the "current bundle or body"?

So without any body file control specifying a namespace 

```
bundle foo
{
  classes:
    "bar";
}
```

Would create `default:bar` or `foo.bar`? I know that vars and classes are a bit special because they are like static public global things which accessible "everywhere".

reference/language-concepts/namespaces.markdown
nickanderson: Yes, good suggestion.

reference/language-concepts/namespaces.markdown
nickanderson: Hummm I don't understand

reference/language-concepts/namespaces.markdown
nickanderson: Yeah, but I was also think to just convert this into a see also reference. There's a relationship I want to point out but I don't want to distract from what this is about.

reference/language-concepts/namespaces.markdown
nickanderson: > by "current namespace" here do you mean "default" or a special magical namespace that exists for the "current bundle or body"?

I mean current as in what this.namespace is at current point in policy.

> So without any body file control specifying a namespace 
> 
> ```
> bundle foo

Missing bundle type

> {
>   classes:
>     "bar";
> }
> ```
> 
> Would create `default:bar` 

Yes.

> or `foo.bar`? 

No, classees have a scope as in lifetime or visibility horizon. That's either bundle or namespace. A bundle scope class only lives or can be seen from that bundle or be undles it calls (if inheritance is explicitly requested).

A single class never has a dot in the name  the presence of a dot indicates a class expression (comparison of multiple classes).

foo.bar indicates a variable, but you said classes:.

> I know that vars and classes are a bit special because they are like static public global things which accessible "everywhere".

reference/language-concepts/namespaces.markdown
craigcomstock: So like `this.bundle` what is the `this` part? Looks like a bundle to me since it is separated by a dot `.` from the variable name `bundle`.

reference/language-concepts/namespaces.markdown
craigcomstock: Ok. so 

```
bundle agent foo
{
  vars:
    "bar" string => "yeah";
  classes:
    "baz";
}
```

Would 
- create `default:foo.bar` available everywhere and always
- create `baz` class but only visible in the foo bundle or bundles that foo calls with `methods` promises?

reference/language-concepts/namespaces.markdown
nickanderson: Reading this again .... I said scoped because i was also referring to how `$(variable)` is == to `$($(this.bundle).variable)` which is == to `$($(this.namespace):$(this.bundle).variable)`

Which I was trying to illustrate in referenced core example: https://github.com/cfengine/core/pull/4966/files#diff-5e851531f57a16834042323a690dbbf9db12336174725839130109d23fa519f7

reference/language-concepts/namespaces.markdown
nickanderson: > Ok. so
> 
> ```
> bundle agent foo
> {
>   vars:
>     "bar" string => "yeah";
>   classes:
>     "baz";
> }
> ```
> 
> Would
> 
> * create `default:foo.bar` available everywhere and always
Correct.
Furthermore, if your inside that bundle (that is to say, `$(this.namespace) == default` AND `$(this.bundle) == foo`)  you could refer to that variable value in one of three ways; `$(bar)`, `$(foo.bar)`, or `$(default:foo.bar)`. If you are in another bundle within the default namespace (that is to say, `$(this.namespace) == default` AND `$(this.bundle) != foo`) you could refer to the variable value in one of two ways; `$(foo.bar)`, or `$(default:foo.bar)`. And, if you are in some other namespace (that is to say, `$(this.namespace) != default`) you can refer to that variable value in only one way; `$(default:foo.bar)`

> * create `baz` class but only visible in the foo bundle or bundles that foo calls with `methods` promises?
Almost Correct.

The `baz` class would be visible inside the foo bundle  (that is to say, `$(this.namespace) == default` AND `$(this.bundle) == foo`) and inside any bundle that you call from that bundle with a `methods` promise which has `inherit => "true"`.

Furthermore, you could refer to that class from within the bundle as either simply `baz` or fully qualified as `default:baz`.

craigcomstock: Yes. All that is excellent explanation. :)

reference/language-concepts/namespaces.markdown
nickanderson: I think that I would normally call the `this` part of `this.bundle` a *bundle scope* or a *context* but maybe those names aren't the best. it's not (well, there should not be any bundle named `this`) and *context* is also a term used to refer to a class or a class expression. It's documented in [Special Variables -> this](https://docs.cfengine.com/docs/3.18/reference-special-variables-this.html).

Presence of a `.` alone doesn't tell you what something is. You have to look at the surrounding context.

For example:
`foo.bar` could be part of a variable reference (`$(foo.bar)`) or it could be part of a class expression (`foo.bar::`). But it could as well be fed to something like `if => "foo.bar"` in which case it's a class expression. Or if it was something like `usebundle => my_bundle('foo.bar')` you would have to go look inside `my_bundle` and determine how it's being used. It could be fed to something like `if => "$(my_bundle_param1)"` in which case it's a class expression, or it could be used with something like `"myvar" string => "$($(my_bundle_param1))"` which would be a variable dereference.

craigcomstock: There is a link there to special variables so that is good. Maybe include one example just so folks get the basic gist without following the link.

PR #2786, opened 2022-07-20
Resized the images dimension on the alert page (master)

chiachenglu:
Resized the images dimension

Ticket: ENT-8877
Changelog: Resized the images dimension


olehermanse (approved):
Thanks!

PR #2788, opened 2022-07-21
ENT-8876: Updated Settings-1.png in the setting page (master)

chiachenglu:
Replaced Settings-1.png image on the setting page.

Ticket: ENT-8876
Changelog: Replaced old images on the setting page


olehermanse:
@chiachenglu In this case, you could have written why in the PR description or commit message. It looks a bit weird since you are reducing the resolution here, but I guess the main reason was to remove the hovered element.

olehermanse (approved):

PR #2841, opened 2022-08-08
Change policy style inside braces

craigcomstock:


getting-started/writing-and-serving-policy/policy-style.markdown
```diff
@@ -17,7 +17,7 @@ guide.
 * vertically align opening and closing curly braces unless on same line
 * promise type = 1 indent
 * context class expression = 2 indents
-* promiser = 3 indents
+* promiser = 3 indents, to allow for adding class guards without changing indent
```
nickanderson, line 20: This part +1

getting-started/writing-and-serving-policy/policy-style.markdown
```diff
@@ -130,12 +131,12 @@ bundle agent example
 {
   vars:
       "people" slist => {
-                          "Obi-Wan Kenobi",
-                          "Luke Skywalker",
-                          "Chewbacca",
-                          "Yoda",
```
nickanderson, line 136: I am not a big fan of this. I find it more difficult to see the difference between the attributes and some list members.

nickanderson (commented):
The note about why 3 indents is good.

Personally, I don't like the suggested changes to list member indentations. I think it makes it more difficult to differentiate between attributes and list members.

getting-started/writing-and-serving-policy/policy-style.markdown
craigcomstock: ah, we should add some attributes here and see how it looks! will do.

craigcomstock: Funny. I went looking at vars docs and even one example has a mix ;) We could run our formatter over our examples `:O`

https://docs.cfengine.com/docs/3.18/reference-promise-types-vars.html#slist

```
    vars:

     "xxx"    slist  => {  "literal1",  "literal2" };
     "xxx1"   slist  => {  "1", @(xxx) }; # interpolated in order
     "yyy"    slist  => {
                        readstringlist(
                                      "/home/mark/tmp/testlist",
                                      "#[a-zA-Z0-9 ]*",
                                      "[^a-zA-Z0-9]",
                                      15,
                                      4000
                                      )
                        };

     "zzz"    slist  => { readstringlist(
        "/home/mark/tmp/testlist2",
        "#[^\n]*",
        ",",
        5,
        4000)
        };
```

craigcomstock: best example seems to be comment or maybe meta for inventory

https://docs.cfengine.com/docs/3.18/reference-promise-types-vars.html#attributes

The goal of this PR is to recommend a "compact style" that is reasonable. When retyping that example for here I see that the braces should be indented according to the attribute indention and the values according to one more level. This is, I think, getting too complex, but this is a style guide so I think that is "OK" as a recommendation. `:)`

```
    vars:

        "stuff" slist => {
            "one",
            "two",
          },
          meta => {
            "inventory",
            "attribute_name=My Stuff
          },
          policy => "free";
```

This "indent twice for list items" is similar to the "indent twice for promiser name" in that it leaves "space" for something that COULD come in later.

Out style and a code formatter, like `black` for python, could choose to collapse lists if they fit inside of 80 character lines.

craigcomstock: Looking at that example makes me think of C, how a prevalent style dictates braces always on a separate line aligned with the "thing above", so we could do that here:

```
bundle agent main
{
  vars:
    "my_val"
      slist =>
      {
        "one",
        "two",
      };

  reports:
    "my_val is $(my_val)";
}
```

Which looks odd but leans towards what I did above with two indents after the initial attribute of `slist`.

craigcomstock: I forgot the extra indent between promise type and promiser. I know the reason behind the extra space but it still feels  a bit "odd". But I suppose lack of fore-thought (might want to add a class guard later) is informing my lean towards simplicity (one indent instead of two until I need them).

olehermanse:
> Personally, I don't like the suggested changes to list member indentations. I think it makes it more difficult to differentiate between attributes and list members.

I disagree. The slist members should be indented 1 level more than the attributes, that's how you differentiate it. Your style makes things go too far to the right, IMO. Things fall off the edge of the editor easily, or outside the code block on website / docs site. Also, with your style, several lines have to be reindented if I change a variable namer or promiser string, that also quickly gets inconsistent.

PR #2876, opened 2022-09-08
Added document headers to search index

aleksandrychev:
Ticket: ENT-9141

### DIFF
--- a/generator/build/search/createIndex.js
+++ b/generator/build/search/createIndex.js
@@ -4,6 +4,10 @@ const {readdir} = require('fs').promises;
 
 const htmlFilesDir = '../../_site';
 
+String.prototype.stripHtmlTags = function () {
+    return this.replace(/<\/?[^>]+(>|$)/g, "");
+}
+
 const getHtmlFiles = async (dir) =>
     await Promise.all(
         (await readdir(dir, {withFileTypes: true}))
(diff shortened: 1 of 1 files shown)


generator/build/search/createIndex.js
```diff
@@ -27,18 +31,23 @@ const getHtmlFiles = async (dir) =>
     const htmlFiles = await getHtmlFiles(htmlFilesDir);
     for (const key in htmlFiles) {
         const htmlContent = fs.readFileSync(`${htmlFilesDir}/${htmlFiles[key]}`).toString();
+
         const htmlContentMatch = htmlContent.match(/<div class="article">([\s\S]*?)<div class="footer-top">/gm);
         const titleMatch = htmlContent.match(/<h1 id="top">([\s\S]*?)<\/h1>/gm);
+        const headersMatch = htmlContent.match(/<h[^>]+>(.*)<\/h[^>]+>/gm);
```
olehermanse, line 37: @aleksandrychev Should the regex require `<h` and then a number? We don't want to match other things like `<head>`, right?

aleksandrychev: good point, changed regex to catch only h[1-6] elements

nickanderson:
let's see what it does.

PR #2877, opened 2022-09-08
Added some simple small examples to files promise type docs

olehermanse:


olehermanse:
@cf-bottom jenkins, please

reference/promise-types/files.markdown
```diff
@@ -7,7 +7,78 @@ tags: [reference, bundle agent, files, promises, files promises, promise types]
...
+    "/tmp/hello"
+      content => "Hello, CFEngine";
+}
+```
+
+Before making changes to the system, CFEngine will check what the current state is, and whether it matches the desired state.
+In this case, CFEngine will create and edit the file only if necessary, if `/tmp/hello` already exists with the correct content, no changes are made.
+
+**Tip:** In this and other examples, we use the `__main__` bundle.
+This allows you to easily run a policy file directly.
```
larsewi, line 27:
```suggestion
This bundle is evaluated when (and only when) running the policy file directly.
```

reference/promise-types/files.markdown
```diff
@@ -7,7 +7,78 @@ tags: [reference, bundle agent, files, promises, files promises, promise types]
...
+bundle agent __main__
+{
+  files:
+    "/tmp/hello"
+      content => "Hello, CFEngine",
+      if => fileexists("/tmp/hello");
+}
+```
+
+In the example above, `fileexists()` is a built in function call.
```
larsewi, line 44:
```suggestion
In the example above, `fileexists()` is a built-in function call.
```

larsewi (approved):
Few suggestions. Otherwise a great improvement in terms of ease-of-use. Thanks!

reference/promise-types/files.markdown
olehermanse: No, my example is easier / more focused on the user, IMO.

reference/promise-types/files.markdown
```diff
@@ -7,7 +7,78 @@ tags: [reference, bundle agent, files, promises, files promises, promise types]
...
+bundle agent __main__
+{
+  files:
+    ubuntu_18|ubuntu_20::
+      "/tmp/hello"
+        create => "true";
+}
+```
+
+The line with `ubuntu_18|ubuntu_20::` is called a class guard - it means the promise will only be evaluated if the `ubuntu_18` or `ubuntu_20` class is defined.
```
nickanderson, line 59: maybe link to glossary, also called class expression

reference/promise-types/files.markdown
```diff
@@ -7,7 +7,78 @@ tags: [reference, bundle agent, files, promises, files promises, promise types]
...
+There are many others available to use, see: [Functions][Functions].
+
+### On specific platforms, ensure a file exists by creating it if necessary
+
+```cf3
+bundle agent __main__
+{
+  files:
+    ubuntu_18|ubuntu_20::
+      "/tmp/hello"
```
nickanderson, line 54: this one doesn't age well. Suggest create a file for NOT ubuntu_16 or something that will keep the output the same over time.

reference/promise-types/files.markdown
```diff
@@ -7,7 +7,78 @@ tags: [reference, bundle agent, files, promises, files promises, promise types]
...
+
+The with attribute and variable expansion allows you to easily store the output of a function in a temporary variable, which you can then expand inside content, or other strings within the same promise:
+
+```cf3
+bundle agent __main__
+{
+  files:
+    "/tmp/hello"
+      content => "Output of uname: $(with)",
+      with => execresult("uname", "useshell");
```
nickanderson, line 75: This doesn't age well, uname output will change necessitating changes to the example. Suggest perhaps something like `stat -c "%U" /` which should always return root.

nickanderson (commented):
I think the inline examples should be replaced with examples from core/examples (which do double duty as tests).

```
[%CFEngine_include_example(<Filename from core/examples>)%]
```

E.g. with these new examples from this PR: https://github.com/cfengine/core/pull/5046

```
[%CFEngine_include_example(files-content.cf)%]
[%CFEngine_include_example(files-content-if-fileexists.cf)%]
[%CFEngine_include_example(files-content-create-class-expression.cf)%]
[%CFEngine_include_example(files-content-with.cf)%]
```

PR #2906, opened 2022-10-03
CFE-4063: Added link targets for variable class expressions and automatic canonification

nickanderson:
Ticket: CFE-4063
Changelog: Title

### DIFF
--- a/reference/language-concepts/classes.markdown
+++ b/reference/language-concepts/classes.markdown
@@ -330,8 +330,19 @@ The `"Monday . Evening"` context will only be true on Monday evenings.
 The `Friday . Evening` context will only be true on Friday evenings.
 See below for more on context operators.
 
-Sometimes it's convenient to put class names in variables. This
-example shows two ways to execute code conditionally based on such
+### Variable class expressions
+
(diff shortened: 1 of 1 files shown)


reference/language-concepts/classes.markdown
```diff
@@ -343,6 +354,7 @@ bundle agent greetings
   reports:
    "$(myclassname)"::
      "Good evening!";
+     "What a wonderful sunset";
```
craigcomstock, line 357: Why is this added here? I don't see outputs from this bundle or an explanation of what this second reports promise "shows" about how variables used in quoted class expressions works. :)

nickanderson: Ah, ill update the output. Good. catch. UGH I HATE INLINE examples.

But, the addition was to speak to this: https://github.com/cfengine/documentation/pull/2906/files#diff-596050329d2a43aa188f8c3d1b06d7aa6d014ca65e3f7dde480e8f992dc807b3R338

> Variables used in quoted class expressions apply to all promises following the
quoted class expression until the end of the promise type or until a new class
expression is stated.

PR #2918, opened 2022-10-24
Fix formatting for the first param in CMDB API docs

vpodzime:
There needs to be a blank line before the first item in the list.


nickanderson:
Yep, backported to 3.18

PR #2944, opened 2022-12-21
Explicitly fallback to first available version in dropdown

Lex-2008:
instead of relying on querySelector - since it's unreliable

### DIFF
--- a/generator/_assets/js/dropdown.js
+++ b/generator/_assets/js/dropdown.js
@@ -13,8 +13,11 @@ document.querySelectorAll('.dropdown-select span').forEach(function (item) {
 });
 
 document.querySelectorAll('.dropdown-select').forEach(function (item) {
+    var selected_version = item.querySelector('a[selected="selected"]');
     // select first version in dropdown if no selected version
     // this happens on build previews, because branch name isn't master there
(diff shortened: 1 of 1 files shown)


Lex-2008:
testbuild: https://ci.cfengine.com/view/Documentation/job/fast-build-and-deploy-docs-master/269/
I wonder if change will be visible at https://docs.cfengine.com/docs/3.21/

generator/_assets/js/dropdown.js
```diff
@@ -13,8 +13,11 @@ document.querySelectorAll('.dropdown-select span').forEach(function (item) {
 });
 
 document.querySelectorAll('.dropdown-select').forEach(function (item) {
+    var selected_version = item.querySelector('a[selected="selected"]');
     // select first version in dropdown if no selected version
     // this happens on build previews, because branch name isn't master there
-    var selected = item.querySelector('a[selected="selected"], a').textContent;
-    item.querySelector('span div').textContent = selected;
+    if (!selected_version) {
+        selected_version = item.querySelector('a[selected="selected"]');
```
aleksandrychev, line 20: here I guess it should be just an 'a' selector

PR #2945, opened 2022-12-22
ENT-9421 Updated supported platforms with the release of 3.21.0

nickanderson:
- Added EL9 hub and agent
- Added Debian 11 hub and agent, x86-64 and arm64
- Added Ubuntu 22.04 hub and agent, x86-64 and arm64
- Removed SLES 11
- Removed Solaris 10 (UltraSparc & x86_64)


nickanderson:
If OK, this needs to be merged, back-ported to 3.21 and doc builds for master and 3.21 triggered.

PR #2977, opened 2023-01-17
CFE-4133: Updated readintlist() docs to leverage macros

nickanderson:
Macros help to keep the documentation consistent and an accurate reflection of
the current state.

Ticket: CFE-4133
Changelog: None


larsewi (approved):
Looks good!

PR #2978, opened 2023-01-18
CFE-4133: Updated readintlist() docs to leverage macros (3.21)

nickanderson:
Macros help to keep the documentation consistent and an accurate reflection of
the current state.


cf-bottom:

PR #2989, opened 2023-02-27
CFE-4158: Clarified that Augments does not expand CFEngine variables

nickanderson:
Ticket: CFE-4158
Changelog: none


nickanderson:
@craigcomstock @olehermanse do you think this is more clear?
Ref question from help list noted in https://northerntech.atlassian.net/browse/CFE-4158

olehermanse (approved):
Makes sense.

PR #3040, opened 2023-05-24
checksum_digests.lmdb was removed in 2014, reflect it in docs

dimaslv:


nickanderson:
please pick back to 3.21 and 3.18

nickanderson:
Actually, never mind. I thought this was from Ole. Thanks for the contribution! I will take care of the rest from here.

PR #3106, opened 2023-11-09
ENT-10432: Added documentation on `const.linesep` variable

larsewi:


reference/special-variables/const.markdown
```diff
@@ -54,6 +66,11 @@ reports:
   "But a string with \n in it does not have a newline!";
 ```
 
+**Note:** The variable `const.endl` is an alias for `const.n` and nothing more.
+It is commonly mistaken to be a platform agnostic line separator. But this has
+never been the case. However, since CFEngine 3.23 we introduced `const.linesep`
```
nickanderson, line 71: I would redact the last 3 sentences.

nickanderson (approved):
Looks ok to me, one suggestion that is ok to take or not.

PR #3151, opened 2023-12-21
Added redirects for /licenses

olehermanse:
This can be useful as a nice shorthand to share in other documents.
Especially if we'd like to move some of this information around
later.


craigcomstock:
commit comment has typos

```
This can be useful as a night shorthand to share in other documents.
Especially if we'd like to move soem of this information around
later.
```
s/night/nice/
s/soem/some/

craigcomstock (approved):

PR #3189, opened 2024-01-10
ENT-10347: Added Debian 12 to supported platforms

nickanderson:
Ticket: ENT-10347
Changelog: None


nickanderson:
@cf-bottom jenkins, please

PR #3219, opened 2024-02-20
ENT-11247: installation.markdown: Added hint for how to spawn VM's

larsewi:
As a first time Digital Ocean user, this little hint would have helped me navigate to the correct place faster.


olehermanse (approved):

mineralsfree (commented):
👍

PR #3263, opened 2024-04-24
examples: Fix missing code block

btriller:
Fix missing code block in [1] due to wrong language identifier (`cf` instead of `cf3`)

[1] https://docs.cfengine.com/docs/master/examples-tutorials-distribute-files-from-a-central-location.html#define-locations


olehermanse (approved):
Thanks!

PR #3264, opened 2024-04-24
Fix some format issues

btriller:
* Ensure unordered lists are rendered
* Do not link to cfengine homepage for cfengine class/file template_method


olehermanse (approved):
Thank you!

I think we should fix the autolinker so that `cfengine` is not automatically linked, but I can do that in another PR :)

olehermanse:
Follow-up: https://github.com/cfengine/documentation/pull/3265

PR #3374, opened 2024-12-19
Add documentation for the env prefix in package promise's options attribute

amousset:
docs with https://github.com/cfengine/masterfiles/pull/2963

### DIFF
--- a/reference/promise-types/packages.markdown
+++ b/reference/promise-types/packages.markdown
@@ -127,6 +127,20 @@ rescue solution when a package module has added functionality which is not
 covered by the package promise API. As such there is no official documentation
 for this attribute, its usage depends on the package module in question.
 
+Some package managers support a special ```env``` prefix to set environment
+variables for the package manager. For example, to set the ```EXTERNAL_URL```
+environment variable:
+
+```cf3
+packages:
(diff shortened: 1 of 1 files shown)


reference/promise-types/packages.markdown
```diff
@@ -656,3 +702,56 @@ bundle agent main
 
 - version `latest` is *not* supported when promising an absence
 - `list-updates` is *not* implemented, snaps are automatically updated by default
+
+### zypper
```
nickanderson, line 706: thank you!

reference/promise-types/packages.markdown
```diff
@@ -441,10 +455,28 @@ bundle agent example
   screen-4.1.0-0.19.20120314git3c2946.el7.x86_64
   ```
 
+* Supports [```options```][packages#options] attribute. Each space separate
```
craigcomstock, line 458: This confused me at first. I thought I might need to add a space between a single list item.

When you say "space separate option" do you mean they will eventually become space separated when added to the command lines used in the modules? I think that's what is happening right?

aka

```
EXTERNALURL=https://example.com DEBUG=1 yum update
```

PR #3403, opened 2025-03-07
ENT-10961, CFE-1840: Added documentation on body fsattrs and the immutable constraint

larsewi:
Ticket: ENT-10961, CFE-1840

### DIFF
--- a/content/reference/promise-types/files/_index.markdown
+++ b/content/reference/promise-types/files/_index.markdown
@@ -2749,6 +2749,59 @@ separator, since the backward slash has a special meaning in a regular
 expression. Literal paths may also use backslash (`\`) as a path
 separator.
 
+### fsattrs
+
+**Description:** Manage file system attributes
+
+**Type** `body fsattrs`
+
+**See also:** [Common body attributes][Promise types#Common body attributes]
+
+**Notes:**
+File system attributes are not supported on all file systems and platforms.
+Hence, CFEngine will do it on a best effort basis (without any guarantees).
+
+**History:**
+* Added in CFEngine 3.27.0
+
+#### immutable
+
+**Description:** Set / clear immutable file system attribute
+
+**Type:** [`boolean`][boolean]
+
+**Example:**
+
+```cf3
+body fsattrs set_immutable
+{
+  immutable => "true";
+}
+```
+
+**Notes:**
+Currently only regular files are supported when configuring the immutable bit.
+Furthermore, this will only work on Linux or BSD based systems with filesystems
+that support it. CFEngine will simply ignore the immutable constraint if it's
+not supported and instead log a verbose message (to avoid too much noise).
+
+- If the immutable constraint is set to `"true"` and the promised file:
+    - **is not** immutable; then the agent will perform all other actions promised
+      before setting the immutable bit.
+    - **is** immutable; then the agent will temporarily clear the immutable bit.
+      The agent will do its best to keep the period of the temporarily cleared bit
+      as short as possible. The immutable bit may be temporarily cleared multiple
+      times during a files promise.
+- If the immutable constraint set to `"false"` and the promised file **is**
+  immutable; then the immutable bit is cleared before performing any other
+  actions promised.
+- If the immutable constraint is unspecified and the promised file **is**
+  immutable; then the agent will fail to do any modifications on the file.
+
+**History:**
+
+* Added in CFEngine 3.27.0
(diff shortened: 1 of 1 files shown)


reference/promise-types/files.markdown
```diff
@@ -2749,6 +2749,45 @@ separator, since the backward slash has a special meaning in a regular
...
+**Description:** Manage file system attributes
+
+**Type** `body fsattrs`
+
+**See also:** [Common body attributes][Promise types#Common body attributes]
+
+**Notes:**
+File system attributes are not supported on all file systems and platforms.
+Hence, CFEngine will do it's best effort without any guarantees.
+
```
nickanderson, line 2763:
```suggestion
**History:**

* Added in CFEngine 3.26.0

```

reference/promise-types/files.markdown
```diff
@@ -2749,6 +2749,45 @@ separator, since the backward slash has a special meaning in a regular
...
+body fsattrs set_immutable
+{
+  immutable => "true";
+}
+```
+
+**Notes:**
+Currently only regular files are supported when configuring the immutable bit.
+Furthermore, this will only work on Linux or BSD based systems with filesystems
+that support it. CFEngine will simply ignore the immutable constrain if it is
```
nickanderson, line 2782: Hopefully there is some log message when it's ignored? Mentioning it might be overboard i guess.

reference/promise-types/files.markdown
```diff
@@ -2749,6 +2749,45 @@ separator, since the backward slash has a special meaning in a regular
...
+not supported.
+
+If the immutable constraint is:
+- set to `"true"`, CFEngine will try to clear the immutable bit (if set) before
+  executing the promise and (re-)set it afterwards.
```
nickanderson, line 2787: But, we won't for example make it false then determine no changes are needed and set it back right? Well only remove that immutable bit temporarily once we know a change is needed.

reference/promise-types/files.markdown
```diff
@@ -2749,6 +2749,45 @@ separator, since the backward slash has a special meaning in a regular
...
+- set to `"false"`, CFEngine will try to clear it before executing the promise.
+- unspecified, CFEngine will not try to clear the immutable bit.
```
nickanderson, line 2789:
```suggestion
- unspecified, CFEngine will not try to clear the immutable bit.

**History:**

* Added in CFEngine 3.26.0

```

nickanderson (changes requested):
We should add history.

reference/promise-types/files.markdown
larsewi: There are some verbose log messages if not supported

reference/promise-types/files.markdown
larsewi: This will be quite difficult to achieve. But I can look into it.

nickanderson: :muscle:

craigcomstock (approved):

craigcomstock (changes requested):
one typo

reference/promise-types/files.markdown
larsewi: Now it should only temporarily clear the immutable bit when needed. And the agent will try to keep the period it's cleared as short as machinely possible

craigcomstock (approved):

reference/promise-types/files.markdown
```diff
@@ -2750,6 +2750,59 @@ separator, since the backward slash has a special meaning in a regular
...
+that support it. CFEngine will simply ignore the immutable constraint if it's
+not supported and instead log a verbose message (to avoid too much noise).
+
+- If the immutable constraint is set to `"true"` and the promised file:
+    - **is not** immutable; then the agent will perform all other actions promised
+      before setting the immutable bit.
+    - **is** immutable; then the agent will temporarily clear the immutable bit.
+      The agent will do its best to keep the period of the temporarily cleared bit
+      as short as possible. The immutable bit may be temporarily cleared multiple
+      times during a files promise.
```
nickanderson, line 2795: What does this mean that it might be cleared multiple times during a files promise? Like cleared, re-applied, cleared, re-applied?

nickanderson (approved):
LGTM, but a question  about what it means to clear the immutable bit multiple times.

olehermanse:
@cf-bottom jenkins, please

PR #3416, opened 2025-03-27
ENT-11325: Added documentation for hostswithgroup() policy function

victormlg:
Merge together:
https://github.com/cfengine/enterprise/pull/834
https://github.com/cfengine/core/pull/5741
https://github.com/cfengine/nova/pull/2403
https://github.com/cfengine/documentation/pull/3416

### DIFF
--- a/reference/functions/hostswithclass.markdown
+++ b/reference/functions/hostswithclass.markdown
@@ -18,7 +18,7 @@ hostnames, ip-addresses or public key SHAs of hosts that have a given class.
 configuration files for (classes of) hosts. Hosts are selected based on the
 classes set during the most recently collected agent run.
 
-[%CFEngine_function_attributes(class, field, hostkey)%]
+[%CFEngine_function_attributes(class, field)%]
 
 **Example:**
 
--- a/reference/functions/hostswithgroup.markdown
+++ b/reference/functions/hostswithgroup.markdown
@@ -1,0 +1,37 @@
+---
+layout: default
+title: hostswithgroup
+published: true
+---
+
+**This function is only available in CFEngine Enterprise.**
+
+[%CFEngine_function_prototype(group_name, field)%]
+
+**Description:** Returns a list from the CFEngine Database with the information `field` of all hosts sharing the group `group_name`.
+
+On CFEngine Enterprise, this hub function can be used to return a list of hostnames, ip-addresses or public key SHAs of hosts that are from the same group. It works very similarly as its twin `hostswithclass()`.
+
+[%CFEngine_function_attributes(group_name, field)%]
+
+**Example:**
+
+```cf3
+bundle agent debian_hosts
+{
+vars:
+
+  am_policy_hub::
+    "host_list" slist => hostswithgroup( "Linux", "name" );
+
+files:
+  am_policy_hub::
+    "/tmp/master_config.cfg"
+         edit_line => insert_lines("host=$(host_list)"),
+            create => "true";
+}
+```
+
+**History:** Was introduced in 3.26.0
+
+**See also:** `hostswithclass()`


reference/functions/hostswithgroup.markdown
```diff
@@ -0,0 +1,39 @@
...
+title: hostswithgroup
+published: true
+---
+
+**This function is only available in CFEngine Enterprise.**
+
+[%CFEngine_function_prototype(group_name, field)%]
+
+**Description:** Returns a list from the CFEngine Database with the information
+`field` of all hosts sharing the group `group_name`.
```
craigcomstock, line 12:
```suggestion
**Description:** Returns a list from the CFEngine Database with the information `field` of all hosts sharing the group `group_name`.
```

might as well go with one-line-per-sentence.

reference/functions/hostswithgroup.markdown
```diff
@@ -0,0 +1,39 @@
...
+
+On CFEngine Enterprise, this function can be used to return a list of
+hostnames, ip-addresses or public key SHAs of hosts that are from the same group. It works very similarly as its twin `hostswithclass()`.
```
craigcomstock, line 15:
```suggestion
On CFEngine Enterprise, this function can be used to return a list of hostnames, ip-addresses or public key SHAs of hosts that are from the same group.
It works very similarly as its twin `hostswithclass()`.
```

reference/functions/hostswithclass.markdown
```diff
@@ -18,7 +18,7 @@ hostnames, ip-addresses or public key SHAs of hosts that have a given class.
 configuration files for (classes of) hosts. Hosts are selected based on the
 classes set during the most recently collected agent run.
 
-[%CFEngine_function_attributes(class, field, hostkey)%]
+[%CFEngine_function_attributes(class, field)%]
```
craigcomstock, line 21: where is the list of options for field?

reference/functions/hostswithgroup.markdown
```diff
@@ -0,0 +1,39 @@
...
+
+[%CFEngine_function_attributes(group_name, field)%]
```
craigcomstock, line 17: Seems like we need mention of possible values for field.
As I understand it would be "name", "ip" or "hostkey"?

reference/functions/hostswithgroup.markdown
nickanderson: I would also note that it is a hub only function.

reference/functions/hostswithgroup.markdown
```diff
@@ -0,0 +1,39 @@
...
+
+**Example:**
```
larsewi, line 19: I think it should be possible to include the example from _core/examples/_ this way we make sure that the example in docs always works.

victormlg: The problem is that if I make an example in docs it would run it and make the acceptance tests fail since this function is only available on enterprise/nova.  So I thought it would make sense to do the same thing as the already existing docs for hostswithclass, and just copy an example there instead of making an example in core/examples/

reference/functions/hostswithclass.markdown
olehermanse: As Nick says, the list of options is what the macro is for, essentially. It gets it from the syntax description (C code).

The change here is correct, the hostswithclass function has 2 arguments, not 3.

reference/functions/hostswithgroup.markdown
olehermanse: This is not a requirement, and shouldn't be. Inline examples are nice. Let's instead work on adding checking for inline examples.

larsewi: Makes sense :+1:

nickanderson (approved):

craigcomstock (approved):

PR #3419, opened 2025-04-09
ENT-12734: Added tool to extract all inline code blocks in CFEngine Docs

victormlg:


cf-bottom:

markdown-code-checker.py
```diff
@@ -0,0 +1,90 @@
...
+
+
+def write_file(file_name, extension, code_snippet):
+    with open(f"{file_name}.{extension}", 'w') as f:
+        f.write(code_snippet)
+
+
+def parse_args():
+    parser = argparse.ArgumentParser(prog="Markdown inline code syntax checker",
+                                     description="checks the syntax of documentation inline code"
```
olehermanse, line 69: In the future we'd want this tool to do a number of things, including autoindentation, syntax checking, maybe even running them.

markdown-code-checker.py
```diff
@@ -0,0 +1,90 @@
...
+                                     )
+    parser.add_argument("--path", "-p", help="path of file or directory to check syntax on", default=".", required=False)
+    parser.add_argument("--languages", "-l", nargs='+', help="languages to check syntax of", default=["cf3", "json", "yaml"], required=False)
```
olehermanse, line 72: Maybe add flags or subcommands for what the command should do, for example:

```
python3 markdown-code-checker.py --path . --languages cf3 --extract
```

And in the future it would be


```
python3 markdown-code-checker.py --path . --languages cf3 --extract --autoformat --syntax-check --replace
```

markdown-code-checker.py
```diff
@@ -0,0 +1,90 @@
+import marko as md
```
olehermanse, line 1: Having this script at the top level is a bit unexpected / noisy. Please put it inside a new `./scripts` folder. (We already do this in some other repos).

markdown-code-checker.py
```diff
@@ -0,0 +1,90 @@
...
+        content = f.read()
+
+    parser = md.parser.Parser()
+    ast = parser.parse(content)
+
+    code_snippet_count = 0
+    for child in ast.children:
+        # TODO: add a way to exclude a code snippet
+        if isinstance(child, md.block.FencedCode) and child.lang in languages:
+            code_snippet_count+=1
```
olehermanse, line 19: In other repos / python tools we run `black` to format the source files. Can you do the same here?

olehermanse (approved):

PR #3421, opened 2025-04-11
 ENT-12735 Added a way to exclude / skip specific code snippets to markdown-code-checker

victormlg:


olehermanse (approved):
Nice!

Next time, try to do the individual changes as separate commits (we fixed most of it now, but the last commit still has 2 things which could be separate - refactor and then change behavior, or change behavior and then refactor).

olehermanse (approved):

PR #3426, opened 2025-04-15
ENT-12737: Added check_syntax for cf3 in markdown parser

victormlg:


scripts/markdown-code-checker.py
```diff
@@ -78,34 +78,39 @@ def check_output():
     pass
 
 
-def replace():
```
olehermanse, line 81: _autofromat_ in commit title

scripts/markdown-code-checker.py
```diff
@@ -78,34 +78,39 @@ def check_output():
...
+        return
+
+    with open(path, "r") as f:
+        markdown_content = f.read()
+        lines = markdown_content.split("\n")
+
+    lines[first_line + 1 : last_line - 1] = pretty_content.split("\n")
+
+    with open(path, "w") as f:
+        f.write("\n".join(lines))
```
olehermanse: All the 3 times you call `open()` an exception can be raised. And the most likely exceptions are file not found, or access denied.

I suggest putting the whole function (all 3 opens) into one big `try...except`, and then `excepting` the specific exceptions like `FileNotFoundError` for nicer error messages.

You can also add the more generic [`IOError`](https://stackoverflow.com/a/70538223) to cover other ways which reading and writing files can fail. Avoid `except` without an exception type, since it can hide other problems we would want to fix, like `AssertionError`, `AttributeError`, etc.

scripts/markdown-code-checker.py
```diff
@@ -73,8 +75,28 @@ def extract(path, i, language, first_line, last_line):
...
+        )
+        return
+
+    match language:
+        case "cf":
+            p = subprocess.run(
+                ["/var/cfengine/bin/cf-promises", abs_file_name],
+                capture_output=True,
+                text=True,
+            )
```
olehermanse: Should add some exception handling here.

scripts/markdown-code-checker.py
```diff
@@ -181,6 +203,14 @@ def parse_args():
         print("[error] This path doesn't exist")
         sys.exit(-1)
 
+    if (
+        args.syntax_check
+        and "cf3" in args.languages
+        and not which("/var/cfengine/bin/cf-promises")
```
olehermanse, line 206: Not a big concern in this tool, but be aware that when you do this (first check if a file / program exists, and then later open it, assuming it still exists) you open up for the possibility of that program being modified or deleted in between. This is called a [Time-of-check to time-of-use race condition](https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use).

Generally, the preferred pattern, especially in Python, is to just try to open it and then handle the exceptions if there's something wrong. Or you can do both, check early as you do here, to give a nice and early error message, but then still handle the exceptions correctly when you open `cf-promises`, because it could have disappeared in the meantime.

olehermanse (dismissed):

scripts/markdown-code-checker.py
```diff
@@ -7,10 +7,10 @@
 import subprocess
 
 
-def extract_inline_code(file_path, languages):
+def extract_inline_code(path, languages):
```
olehermanse, line 10: This commit is not very nice, here is the commit message:

```
Changed var names, parameters and error handling
```

It would be much nicer to review if it was only changing the names and leaving the error handling for a separate commit. Also, in this commit you are removing the error handling (try except) only to reintroduce it / fix it in the next commit. Let's take a look at how to fix this together when you have time.

olehermanse (approved):

PR #3430, opened 2025-05-08
CFE-2815: Added documentation for `isnewerthantime`

jakub-nt:
corresponding core PR: https://github.com/cfengine/core/pull/5799

### DIFF
--- a/reference/functions/isnewerthantime.markdown
+++ b/reference/functions/isnewerthantime.markdown
@@ -9,8 +9,9 @@ published: true
 **Description:** Returns whether the file `file` is newer (modified later)
 than the time `time`.
 
-This function compares the modification time (mtime) of the file, referring to
-changes of content only.
+This function compares the modification time (mtime) of the file. Note that
+access changes such as ownership and permissions as well as status changes
+such as last time the file was read are not included in the mtime timestamp.
 
 [%CFEngine_function_attributes(file, time)%]
 
--- a/reference/functions/isnewerthan.markdown
+++ b/reference/functions/isnewerthan.markdown
@@ -9,8 +9,9 @@ published: true
 **Description:** Returns whether the file `newer` is newer (modified later)
 than the file `older`.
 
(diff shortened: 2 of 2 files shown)


reference/functions/isnewerthantime.markdown
```diff
@@ -0,0 +1,31 @@
...
+published: true
+---
+
+[%CFEngine_function_prototype(file, time)%]
+
+**Description:** Returns whether the file `file` is newer (modified later)
+than the time `time`.
+
+This function compares the modification time (mtime) of the files, referring
+to changes of content only.
```
craigcomstock, line 13: The "referring to changes of content only" doesn't make sense to me. Doesn't isnewerthantime() just check the mtime and that's it?

craigcomstock (changes requested):

reference/functions/isnewerthantime.markdown
jakub-nt: Yes, it just checks the `mtime`, and the point of this sentence (which is copy pasted from `isnewerthan` docs) is to clarify that it doesn't compare other data e.g. attributes (e.g. permissions) of the file.

reference/functions/isnewerthantime.markdown
```diff
@@ -0,0 +1,31 @@
...
+published: true
+---
+
+[%CFEngine_function_prototype(file, time)%]
+
+**Description:** Returns whether the file `file` is newer (modified later)
+than the time `time`.
+
+This function compares the modification time (mtime) of the file, referring to
+changes of content only.
```
craigcomstock, line 13:
```suggestion
This function compares the modification time (mtime) of the file. Note that access changes such as ownership and permissions as well as status changes such as last time the file was read are not included in the mtime timestamp.
```

craigcomstock (commented):
This is how I would word it so folks don't have to lookup mtime for specificity.

craigcomstock (approved):

PR #3453, opened 2025-07-01
functions: Remove quotes from link text

btriller:


olehermanse (approved):
Thanks, but we should fix this in the site generator as well.

https://northerntech.atlassian.net/browse/CFE-4552

PR #3496, opened 2025-08-26
Fixed missing rvm install in docs-revamp-22 container images on new build hosts

craigcomstock:
When we removed rvm install in generator/build/Dockerfile we needed to change the name of the buildah image so that there were two: one using ruby/rvm for pre-hugo and one for hugo going forward.

This mostly affected newer build hosts like elfk-spectre.

Ticket: ENT-13085
Changelog: none

### DIFF
--- a/generator/build/run.sh
+++ b/generator/build/run.sh
@@ -3,8 +3,8 @@
 set -ex
 trap "echo FAILURE" ERR
 
-if ! buildah inspect docs-revamp-22 >/dev/null 2>&1; then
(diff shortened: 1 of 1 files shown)


craigcomstock:
nope. this failed. dependency hell is here. I copied a working container image from an old build host and that works for now.
 the todo here is to automate the generation of a golden docs container image on a regular basis so we see problems sooner and can maintain the setup from "scratch". See ENT-13193 for this work.

PR #3506, opened 2025-09-12
methods: Fix inherit example

btriller:
Also fix a typo


cf-bottom:

PR #3507, opened 2025-09-12
Include common attributes for promise types

btriller:


nickanderson:
@cf-bottom jenkins, please

PR #3634, opened 2026-04-23
Added nightly doc build triggers

nickanderson:


craigcomstock (approved):
doesn't really make sense to me completely but let's give it a try.

Jenkinsfile
```diff
@@ -20,14 +20,83 @@ pipeline {
...
+      when {
+        triggeredBy 'TimerTrigger'
+      }
+      steps {
+        script {
+          echo "Triggered by cron - launching builds for master, 3.27, 3.24, and lts"
+
+          // Build master
+          build job: env.JOB_NAME, parameters: [
+            string(name: 'CORE_REV', value: 'master'),
```
craigcomstock, line 40: would be nice to parameterize this a bit so not so much repetition, but there are differences here and there.

Jenkinsfile
```diff
@@ -20,14 +20,83 @@ pipeline {
...
+            string(name: 'PACKAGE_JOB', value: 'cf-remote'),
+            string(name: 'USE_NIGHTLIES_FOR', value: '3.27.x')
+          ], wait: false
+
+          echo "All nightly builds triggered successfully"
+        }
+      }
+    }
     stage('Environment check') {
+      when {
```
craigcomstock, line 95: why skip these?

nickanderson: I think so that a manual trigger doesn't cause all of them to build this one cron trigger causes master, 3.27, 3.24 and lts (which is 3.27 but targeting /lts ) to build.

PR #3724, opened 2026-07-06
CFE-4694: Updated docs for string_replace

victormlg:
https://github.com/cfengine/core/pull/6219

### DIFF
--- a/content/reference/functions/string_replace.markdown
+++ b/content/reference/functions/string_replace.markdown
@@ -5,14 +5,15 @@ aliases:
   - "/reference-functions-string_replace.html"
 ---
 
-{{< CFEngine_function_prototype(string, match, replacement) >}}
+{{< CFEngine_function_prototype(string, match, replacement, option) >}}
 
-**Description:** In a given string, replaces a substring with another string.
-
-{{< CFEngine_function_attributes(string, match, replacement) >}}
+**Description:** Replaces a substring, or a list of substrings, within a given string with a specified replacement string or list of replacement strings.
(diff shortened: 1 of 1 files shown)


content/reference/functions/string_replace.markdown
```diff
@@ -7,18 +7,19 @@ aliases:
...
 {{< CFEngine_function_prototype(string, match, replacement) >}}
 
-**Description:** In a given string, replaces a substring with another string.
-
+**Description:** Replaces a substring, or a list of substrings, within a given string with a specified replacement string or list of replacement strings.
 {{< CFEngine_function_attributes(string, match, replacement) >}}
 
 Reads a string from left to right, replacing the occurences of the second
-argument with the third argument in order.
+argument with the third argument in order. When lists are provided for both match and replacement, they must be of equal length, and substitutions are applied pairwise.
```
nickanderson: So, the agent will error if you pass it a list and a string or a string and a list or lists with lengths that do not match?

victormlg: I changed a bit how it works to ensure backwards compatibility. If the user wants to replace using one single match, they can write:

`string_replace("input", "input", "output", "strings") ` or simply  `string_replace("input", "input", "output") `

If they want to replace using lists of strings, then yes, the types must match and their length must match. Also it needs to be specified with `"lists"`:

`string_replace("input", "@(matches)", "@(replacements)", "lists") `. Now they can also write inline json: `string_replace("input", '["i", "p"]', '["I", "P"]', "lists")`

nickanderson (approved):

PR #3742, opened 2026-07-21
CFE-3725: hash=>"best" resolves to SHA-512 (community)

nickanderson:
## Summary

Companion to cfengine/core#6241.

`hash => "best"` now resolves to a single **SHA-512** instead of cross-correlating MD5 and SHA1. This updates the `changes` body `hash` attribute reference:

- Description: "cross correlates the best two available algorithms known in the OpenSSL library" → "uses the strongest hash algorithm available (SHA-512)."
- Adds a **History** note: the change landed in **3.29.0** and affects the **community edition only** (the enterprise edition already used SHA-512). Files tracked with `hash => "best"` are silently re-baselined under the new hash on the first agent run after upgrade.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/content/reference/promise-types/files/_index.markdown
+++ b/content/reference/promise-types/files/_index.markdown
@@ -640,7 +640,7 @@ body acl template
 
 **Description:** The `hash` menu option defines the type of hash used for change detection.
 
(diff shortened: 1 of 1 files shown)


cf-bottom:

PR #3764, opened 2026-09-03
CFE-3275: Documented where report_diffs stores its cached copies

nickanderson:
Documents the `$(sys.statedir)/change_details` location that `report_diffs` cached copies moved to in CFE-3275.


cf-bottom:

